PR #2, opened 2019-04-01
use regional subnets for bastion, group all bastion resources, remove…

hyder:
…d ksonnet, updated docs


modules/base/bastion/scripts/bastion.template.sh
```diff
@@ -1,5 +1,177 @@
-#!/bin/bash
 # Copyright 2017, 2019, Oracle Corporation and/or affiliates.  All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl
 
+#!/bin/bash
```
markxnelson: shebang line has to be first line

PR #17, opened 2019-05-03
Adapt module to make it more capable of provisioning multiple clusters

cawilliamson:
The idea here is that the kubeconfig file will be named based on the cluster itself since when trying to provision multiple clusters in the same TF setup I anticipate you would experience clashes between the various kubeconfig files with the original implementation.

This includes a few related changes:

- Replace kubeconfig output with kubeconfig_env (current behaviour) and kubeconfig_file (basically the same as previous but without the 'export KUBECONFIG=' stuff.)

- Added variable for config_output_path (in-line with what the AWS module uses) and defaulted that to generated/ (in-line with your current implementation.)

- Replaced all hard-coded references to kubeconfig with the new dynamic filenaming strategy.

- Adds output value of the API endpoint (incl. HTTPS)

These changes combined should allow k8s clusters to be easily provisioned and then instantly used within a single TF deployment.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -37,6 +37,9 @@ module "base" {
   image_operating_system         = "${var.image_operating_system}"
   image_operating_system_version = "${var.image_operating_system_version}"
 
+  # kubeconfig
+  config_output_path   = "${var.config_output_path}"
+
   # availability_domains
   availability_domains = "${var.availability_domains}"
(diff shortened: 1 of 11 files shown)


cawilliamson:
Discarding this PR in favour of the other.

Issue #18, opened 2019-05-03
[Bug] Helm is needed when it shouldn't be


cawilliamson:
In your latest commit you seem to have broken the helm dependency.

I have no intentions of using Helm and am using the default of install_helm=false *but* the binary itself is now needed since the ~/.helm dir needs to exist in order for your TF code to execute.

hyder:
Sorry about that. I actually misunderstood how the helm provider works. Also, I found an open bug on the helm provider yesterday. I'll look at your fix

hyder:
Fixed in https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/29

PR #23, opened 2019-05-07
Removed hard dependency on helm

cawilliamson:
This change removes the hard dependency on Helm which was there since the tooling cleanup.

It resolves issue #18.

### DIFF
--- a/modules/oke/helm.tf
+++ b/modules/oke/helm.tf
@@ -10,6 +10,8 @@ provider "helm" {
 data "helm_repository" "stable" {
   name = "stable"
   url  = "https://kubernetes-charts.storage.googleapis.com"
+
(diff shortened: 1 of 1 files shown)


hyder:
Sorry about that. I actually misunderstood how the helm provider works. Also, I found an open bug on the helm provider yesterday. 

I'm removing this whole bit altogether as there's no need for them here:

provider "helm" {
  kubernetes {
    config_path = "${path.root}/generated/kubeconfig"
  }
}

data "helm_repository" "stable" {
  name = "stable"
  url  = "https://kubernetes-charts.storage.googleapis.com"
}

Do you mind testing the prom-operator branch (which I've just pushed)? You can ignore the prometheus-operator stuff.

PR #24, opened 2019-05-07
Bring remaining variables in-line with tfvars example file

cawilliamson:


hyder:
Thanks for raising this. We actually need to sync the values of the documentation, the tfvars example and the root variables.tf. I've done this based on your branch and will create a new PR to merge. I'll close this for now

PR #25, opened 2019-05-07
Remove demo contents

cawilliamson:
The purpose of this PR is ultimately to try to get this module production ready.

There are a few frustrations right now which this addresses:

- Duplication of effort between maintaining variables and terraform.tfvars.example (which other than as a demo - I don't see any value in?)

- Duplication of parameters when using the OCI provider within a TF script. ALL already set OCI provider parameters are overridden by this individual module which doesn't make it suitable for use in an actual terraform script.


hyder:
It's helpful for those who are getting started.

PR #39, opened 2019-07-06
Update nodepools.tf to remove the hardcoded image name

karthicgit:
Image name has been hardcoded in variable file earlier. Oracle will release latest OS and instead of hardcoding the latest image name ,code has been changed to fetch the latest image id.

Signed-off-by: Karthicgit karthic.ravindran@oracle.com


hyder:
Hi, thanks for contributing. I'll include this PR and then move it to the base module

PR #83, opened 2019-10-09
Instructions for git clone was referring to old repo.

ashageetha:
Instructions for git clone was referring to old repo. Updated to point to current repository.


markxnelson:
Thank you very much @ashageetha

PR #93, opened 2019-11-14
Issue 86 - Updated bastion-instance-principal policy to use compartment id

kumar-dhanagopal:
Updated bastion-instance-principal policy to use compartment id instead of compartment name, and removed the compartment_name variable.


hyder:
already fixed in #94. can you please test the 91-admin-host branch?

PR #161, opened 2020-04-28
Orace Vault Secret to store authtoken for accessing OCIR

karthicgit:
Authtoken will be stored in Oracle vault secret and this will be used as a imagepullsecrets in Kubernetes to fetch image from OCIR.


livespotty:
This PR has been tested on Montréal Region. Also tested using oracle keyvault to reuse ocir tokens. There is no issue found and works as expected. Thank you.

PR #167, opened 2020-06-08
Issue 164

karthicgit:
Added a new variable check_node_active to wait till worker nodes are active.Changed default value of secret_id to NONE.


modules/oke/scripts/is_worker_active.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/python
+#!/usr/bin/python3
```
hyder, line 1: This will check if there is 1 active node. I think the request was to check for all nodes being active. 

Can we do the following:

1. change the variable to string
2. the possible values should be check_none, check_one, check_all
3. execute the script anyway but change its behaviour according to the value passed

?

karthicgit: Closing this pull request and raising another one with the suggested changes.

Issue #164, opened 2020-05-21
would be possible to implement a "waiting state" until all the worker nodes are ready ?


eugsim1:
Hi I m using your excellent implementation, but i m hitting a feature of the terraform kubernetes call.
Whenever i want to modify the pool size expand/reduce the request is executed immediately ie the terraform scripts are setting up the new pool size, but the creation of the nodes is taking time and the new configuration is not ready yet meanwhile the terraform script has finished the excution.
If i have another scirpt which has to run on the new pool it will fail as the nodes are not ready yet.
would be possible to implement a flag like "keep_pooling_until_all_nodes_are_ready" so the execution of the terraform script will finish when the pool is having the new settings ?
thanks
eugene

hyder:
done in #174

PR #168, opened 2020-06-09
Check till worker nodes are active related to Issue #164.

karthicgit:
check_node_active variable has been added . Option to check only one worker node is active or all the worker nodes are active.


docs/terraformoptions.adoc
```diff
@@ -432,6 +432,11 @@ tags = {
 |
 |
 
+|check_node_active
+|Setting this variable to *one* will wait till any one worker node is active.Setting to *all* will wait till all worker node is active.*NONE* means no check.
+|"one","all","NONE"
```
hyder, line 437: Make possible values standard i.e. all in caps or without caps. Verify what we've done where we've used none before and keep it the same please.

Issue #176, opened 2020-07-17
node_pool_image_id shoud be "NONE"


kranthikirang:
Getting the node_pools latest_images will fail if the node_pool_image_id="none"

In modules/oke/nodepools.tf code as below
`node_source_details {
    image_id    = var.node_pools.node_pool_image_id == "NONE" ? data.oci_core_images.latest_images[count.index].images[0].id : var.node_pools.node_pool_image_id
    source_type = "IMAGE"
  }`

Error will be like below

`module.oke.oci_containerengine_node_pool.nodepools[0]: Creating...

Error: Service error:InvalidParameter. Unable to get image information for the node pool.. http status code: 400. Opc request id: 470ee5cb155b85dd9c3b2a2eeef0b7d6/6871BF8AC060CDDD6AF36022D41F533A/16DA21168ADC48C8A8DAD303E7E69946

  on modules/oke/nodepools.tf line 4, in resource "oci_containerengine_node_pool" "nodepools":
   4: resource "oci_containerengine_node_pool" "nodepools" {`

hyder:
Hi,

Thanks for logging this request. The code you highlighted above shows the following:

```
image_id    = var.node_pools.node_pool_image_id == "none" ? data.oci_core_images.latest_images[count.index].images[0].id : var.node_pools.node_pool_image_id
```

and therefore the value in the example file should also be in lower case.

Can you please confirm you are using the latest?

Thanks.

kranthikirang:
oh my bad. Looks like I was using v2.2.2 and this issue was addressed in master branch. Closing then

PR #177, opened 2020-07-17
node_pool_image_id changed as NONE

kranthikirang:
* Issue 176 fix.
* In order to properly get the list of latest images we need to pass node_pool_image_id as "NONE" instead of "none"

Signed-off-by: kranthi guttikonda <kranthi.guttikonda@b-yond.com>


hyder:
Resolves #176

PR #205, opened 2020-08-24
Allow public ingress from anywhere to ports 80 and 443

### LINKED ISSUE #188: LB subnet shouldn't allow traffic from internet on all ports
Load Balancer subnet allows traffic from internet on all ports - Can this be addressed? Would be better to only allow 80 and 443 as default. 

Thanks for looking into this. You all have delivered an awesome solution.

noeldcosta:
Closes #188 

Signed-off-by: Noel Dcosta <noel.dcosta@oracle.com>

### DIFF
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -554,6 +554,11 @@ Refer to {uri-topology}[topology] for more thorough examples.
 |internal/public
 |public
 
+|`public_lb_ports`
+|The list of destination ports to allow for public ingress.
+|List of ports.
+|80, 443
+
 |`waf_enabled`
 |Whether to enable WAF monitoring and protection of public load balancers.
 |true/false
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -205,7 +205,7 @@ newbits = {
 
 |`bastion_access`
 |CIDR block in the form of a string to which ssh access to the bastion must be restricted to. *_ANYWHERE_* is equivalent to 0.0.0.0/0 and allows ssh access from anywhere.
-|XXX.XXX.XXX.XXX/YY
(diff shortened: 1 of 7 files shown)


hyder:
Instead of hardcoding 80 & 443, what do you think of setting these 2 as default but let users to specify their own list of ports?

+ @markxnelson

noeldcosta:
have a look at https://github.com/oracle-terraform-modules/terraform-oci-oke/compare/master...noeldcosta:188-fixes

noeldcosta:
@hyder the change set has been updated with support for a custom list of ports

hyder (changes requested):
Please also document the changes in:
1. Configuration (under Configure OKE Load Balancer parameters)
2. Terraform Options

Lastly, add them in the terraform.tfvars.example file

noeldcosta:
Documentation has been updated.

docs/terraformoptions.adoc
```diff
@@ -554,6 +554,11 @@ Refer to {uri-topology}[topology] for more thorough examples.
 |internal/public
 |public
 
+|`public_lb_ports`
+|The list of destination ports to allow for public ingress.
+|List of ports.
```
hyder, line 559: 3rd column should be an example [80,443]

modules/okenetwork/variables.tf
```diff
@@ -35,8 +35,9 @@ variable "lb_subnet_type" {
   type = string
 }
 
+variable "public_lb_ports" {}
```
hyder, line 38: please set type list(string)

noeldcosta: Have set it to below to match the variable definition in variables.tf

variable "public_lb_ports" {
  type = list(number)
}

hyder (approved):

PR #214, opened 2020-09-02
reversed 2.3.2 and configure aliased providers fully

hyder:


hyder:
@karthicgit

karthicgit:
Tested the PR with below usecase
1.user configure provider.tf and no oci-cli configured.The values given in provider.tf will be picked .
2.user has oci-cli configured and no provider.tf is used. The configuration mention in oci-cli will take preference .For example if the user has mentioned region as ca-toronto-1 in tf.vars but oci-cli config has region as us-phoenix-1 the resources will get created in pheonix.

hyder:
Thanks. We should make sure that we don't depend on the oci-cli config for provider config.

PR #217, opened 2020-09-08
PR for issue-216 add support for flex shape

karthicgit:
Signed-off-by: Karthic Ravindran <karthic.ravindran@oracle.com>
In nodepool map variable an extra parameter has been added for Flex shape OCPU.

### DIFF
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -456,22 +456,35 @@ admission_controller_options = {
 |v1.16.8
 
 |`node_pools`
-|The number, shape and quantities per subnets of node pools to create. Each key and tuple pair corresponds to 1 node pool. The first parameter in the tuple sets the shape of the worker node, the 2nd parameter sets the size of the node pool and the 3rd parameter the boot volume size in GB. An empty value ("") for the boot volume size will default the boot volume size to 50GB. A minimum of 1 worker worker nodes per node pool will be created.  Refer to {uri-topology}[topology] for more thorough examples.
+a|The number, shape of node pools and node_pool_size to create. Each key and tuple pair corresponds to 1 node pool.
+
+* shape defines the worker node shape to use for each pool
+* ocpus defines the number of OCPUs that will be used if VM.Standard.E3.Flex shape is used
+* node_pool_size defines the number of worker nodes in each nodepool
+* boot_volume_size defines the custom boot volume size in GBs for the worker nodes.
+
+If an empty nodepool like np3 = {} is specified, then a nodepool will default values: 
+
+* shape=VM.Standard.E3.Flex
+* ocpus=1
+* node_pool_size=1
+* boot_volume_size=50
+
+Refer to {uri-topology}[topology] for more thorough examples.
 |e.g.
 [source]
 node_pools = {
-  np1 = ["VM.Standard.E2.2", 1, 50]
-  np2 = ["VM.Standard2.8", 2, 100]
-  np3 = ["VM.Standard.E2.2", 1, 200]
-  np4 = ["VM.Standard.E2.2", 1, ""] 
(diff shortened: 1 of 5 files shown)


docs/terraformoptions.adoc
```diff
@@ -456,22 +456,22 @@ admission_controller_options = {
 |v1.16.8
 
 |`node_pools`
-|The number, shape and quantities per subnets of node pools to create. Each key and tuple pair corresponds to 1 node pool. The first parameter in the tuple sets the shape of the worker node, the 2nd parameter sets the size of the node pool and the 3rd parameter the boot volume size in GB. An empty value ("") for the boot volume size will default the boot volume size to 50GB. A minimum of 1 worker worker nodes per node pool will be created.  Refer to {uri-topology}[topology] for more thorough examples.
+|The number, shape and quantities per subnets of node pools to create. Each key and tuple pair corresponds to 1 node pool. The first parameter in the tuple sets the shape of the worker node,the 2nd parameter sets the number of OCPU for flex shape compute and minimum one OCPU will be used. Its set to null for standard shapes, the 3rd parameter sets the size of the node pool and the 4th parameter the boot volume size in GB. An empty value ("") for the boot volume size will default the boot volume size to 50GB. A minimum of 1 worker worker nodes per node pool will be created.  Refer to {uri-topology}[topology] for more thorough examples.
```
hyder, line 459: We don't use quantities per subnets for node pools anymore. We use cluster size

hyder: don't set to null for standard shapes. just ignore it altogether. see if you can add dynamic before the node_shape_config.

hyder: Also add that the ocpus is for each worker node.

modules/oke/nodepools.tf
```diff
@@ -21,11 +21,15 @@ resource "oci_containerengine_node_pool" "nodepools" {
       }
     }
     # set quantity to a minimum of 1 to allow small clusters. 
-    size = max(1, element(each.value,1))
+    size = max(1, element(each.value,2))
   }
+  node_shape_config {
```
hyder, line 26: see if you can use a dynamic here like thus: https://github.com/oracle-terraform-modules/terraform-oci-vcn/blob/master/nat.tf#L27

so the node_shape_config will only be evaluated iff the shape is flex.

variables.tf
```diff
@@ -282,9 +282,9 @@ variable "kubernetes_version" {
...
 variable "node_pools" {
   default = {
-    np1 = ["VM.Standard.E2.2", 1, 50]
-    np2 = ["VM.Standard2.8", 2, 100]
-    np3 = ["VM.Standard.E2.2", 1, 200]
+    np1 = ["VM.Standard.E2.2",null, 1, 50]
+    np2 = ["VM.Standard.E3.Flex",3, 2, 100]
+    np3 = ["VM.Standard.E2.2",null, 1, 200]
   }
   description = "Tuple of node pools. Each key maps to a node pool. Each value is a tuple of shape (string) and size(number)."
```
hyder, line 289: Update the description with the order of the parameters

hyder (changes requested):
Suggested changes in docs and code.

docs/terraformoptions.adoc
karthicgit: > don't set to null for standard shapes. just ignore it altogether. see if you can add dynamic before the node_shape_config.

https://docs.cloud.oracle.com/en-us/iaas/api/#/en/iaas/20160918/datatypes/ShapeOcpuOptions
ocpus is a number type.Its better to pass null as we are assigning the same in the code as well.

karthicgit:
@hyder Changes made for Flex shape and OCIR region for San Jose has been added in variables.tf file. Please verify.

hyder (approved):

PR #240, opened 2020-11-20
Changes for adding flex memory in node pool

karthicgit:


hyder (approved):
Verified. Working perfectly.

PR #243, opened 2020-12-03
updated default kubernetes version to v1.18.10, fixed deprecated inte…

hyder:
…rpolation-only expressions


karthicgit (approved):
Approved.

PR #255, opened 2020-12-21
Base module to 2.0.0

hyder:
- Base module to 2.0.0
- Simplified and upgraded calico installation
- Added support for VPA


karthicgit (approved):
Looks good. Tested with metric server,vpa and calico enabled and it ran fine.

PR #264, opened 2021-02-02
Egress rule to allow public lb to reach internal lb

hyder:


hyder:
https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/266

PR #271, opened 2021-03-18
Added support for VCN Native endpoint

hyder:
* Simplified security lists                                                                                                                                                                   
*  Added a subnet for control plane                                                                                                                                                    
* Added 2 parameters                                                                                                                                                                          
* Updated default kubernetes version                                                                                                                                                     
* Updated terraform options doc


hyder:
Test cases:


endpoint | worker | lb
-- | -- | --
public | public | public
public | private | public
public | private | private
public | private | both
  |   |  
private | public | public
private | private | public
private | private | private
private | private | both

kral2:
I would suggest [using imperative mood](https://chris.beams.io/posts/git-commit/#imperative) for issue and commit messages:

- added -> add
- simplified -> simplify

That will help in the long run.

modules/okenetwork/outputs.tf
```diff
@@ -3,6 +3,7 @@
 
 output "subnet_ids" {
   value = map(
+    "cp", join(",", oci_core_subnet.cp.*.id),
```
kral2, line 6: This splat expression syntax is legacy (attribute-only). You should switch to the current splat expression syntax.

```suggestion
    "cp", join(",", oci_core_subnet.cp[*].id),
```

Same suggestion from line 6 to 9.

kral2 (changes requested):

kral2 (dismissed):

PR #278, opened 2021-04-12
Enforce signed images

karthicgit:
To enforce signed images in OKE.

### DIFF
--- a/modules/oke/cluster.tf
+++ b/modules/oke/cluster.tf
@@ -15,7 +15,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   image_policy_config {
     is_policy_enabled = var.oke_cluster.use_signed_images
     key_details {
-      kms_key_id = var.oke_cluster.use_signed_images == true ? var.oke_cluster.kms_key_id : null
+      kms_key_id = var.oke_cluster.use_signed_images == true ? var.oke_cluster.signedimage_key_id : null
     }
   }
 
--- a/modules/oke/cluster.tf
+++ b/modules/oke/cluster.tf
@@ -12,6 +12,13 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
     subnet_id            = var.oke_cluster.cluster_subnets["cp"]
   }
 
+  image_policy_config {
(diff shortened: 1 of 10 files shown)


modules/oke/cluster.tf
```diff
@@ -12,6 +12,13 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
     subnet_id            = var.oke_cluster.cluster_subnets["cp"]
   }
 
+  image_policy_config {
+    is_policy_enabled = var.oke_cluster.use_signed_images
+    key_details {
+      kms_key_id = var.oke_cluster.use_signed_images == true ? var.oke_cluster.kms_key_id : null
```
hyder, line 18: The key used for signed images can be different from the one used to encrypt etcd, so users must be able to specify each separately. 

Add 1 more variable to allow the user to specify the key for signed images and then pass it along to the oke module and eventually to this line of code.

hyder (changes requested):

modules/oke/cluster.tf
hyder: Please add documentation for the new variable in terraform options as well.

karthicgit: New variable signedimage_key_id has been added. Doc has been updated as well.

hyder: 1. Please rename to image_signing_keys.
2. Sorry, I just noticed this: You can add multiple keys for image signing. So, this should be a list of key ids.
3. Add a dynamic and iterator so you can iterate over the list of key ids

karthicgit: Looks like it does not support list. Inappropriate value for attribute "kms_key_id": string required.

hyder: Try putting a dynamic at key_details level  and then add an iterator?

karthicgit: Changes done please review. I have updated the default node_pool_os_version to 7.9 in tfvars.example as well.

hyder (approved):

hyder (dismissed):
Approving review with incoming changes.

hyder (approved):

PR #282, opened 2021-04-16
Adding Initial node label

karthicgit:
To add initial node label for nodepools. The nodepool variable type has been changed to any to accommodate string and map .


karthicgit:
Creating another PR with proper commits.

PR #285, opened 2021-04-19
Bug fix for deprecated map function

hyder:


karthicgit:
You have already converted into a map . So tomap conversion is not required.

karthicgit (approved):

PR #302, opened 2021-05-24
node_pool_image_id changes for GPU

karthicgit:
Signed-off-by: Karthic Ravindran <karthic.ravindran@oracle.com>
This pull request includes changes for node_pool_image for GPU shape included. And also added default values  of [] for image_signing_key variable.


hyder:
Please also test against ARM shapes. I think you might need to modify: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/oke/nodepools.tf#L27

hyder (approved):
Approving for now to get around the failed images lookup. We need to find a better and more stable way of using images. The current dynamic lookup option is not working. It constantly breaks and with ARM and GPU, it will keep happening more often

12345ieee:
Hi guys, I'm really looking forward to this being released, as I have multiple clusters whose terraform apply now fails unless I set the image manually, and then keep it updated manually as well.

Any chance we can get a v3.2.1 soon-ish?

hyder:
Hi @12345ieee 

Can you please describe the problem you are having now?

12345ieee:
Oh sorry, maybe I wasn't clear.
I'm having the issue this PR resolves, which is documented across multiple issues.

I was wondering if you have a timeline to get a version with this fix pushed to the terraform registry.

PR #307, opened 2021-06-25
fix: using correct file to install calico for policy only

hyder:
Changed back to using calico-policy-only.yaml to install calico.

There's a difference between the current version (3.19) and when we were last using (3.12):

The CALICO_IPV4POOL_CIDR and default pod range is now commented, so we added a couple of extra steps to set it as before.


modules/oke/calico.tf
```diff
@@ -24,16 +31,16 @@ resource null_resource "calico_enabled" {
...
-      "$HOME/calico_enabled.sh",
-      "rm -f $HOME/calico_enabled.sh"
+      "chmod +x $HOME/install_calico.sh",
+      "$HOME/install_calico.sh",
+      # "rm -f $HOME/install_calico.sh"
     ]
   }
 
-  count = local.post_provisioning_ops == true && var.calico_enabled == true ? 1 : 0
+  count = local.post_provisioning_ops == true && var.calico.install_calico == true ? 1 : 0 ? 1 : 0
```
karthicgit, line 45: Typo error .

karthicgit (approved):

PR #323, opened 2021-08-19
[Feature] make OCIR secret namespace configurable

### LINKED ISSUE #318: [Feature] Specify OCIR secret namespace
At this moment, the OCIR secret is created under the "default" namespace. I'd like to have an opportunity to specify the namespace where it has to be created.

Use case: keep all resources related to a particular application within the same namespace.

denismakogon:
Closes #318

### DIFF
--- a/modules/oke/scripts/secret.py
+++ b/modules/oke/scripts/secret.py
@@ -15,7 +15,7 @@ secret_id       = '${secret_id}'
 secret_name     = '${secret_name}'
 tenancy_namespace    = '${tenancy_namespace}'
 username        = '${username}'
-namespace       = '${namespace}'
+namespace       = '${secret_ns}'
 
 signer = oci.auth.signers.InstancePrincipalsSecurityTokenSigner()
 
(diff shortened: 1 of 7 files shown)


modules/oke/scripts/secret.py
```diff
@@ -15,6 +15,7 @@
 secret_name     = '${secret_name}'
 tenancy_namespace    = '${tenancy_namespace}'
 username        = '${username}'
+namespace       = '${namespace}'
```
karthicgit, line 18: could you please change the name to secret_ns.

modules/oke/scripts/secret.py
```diff
@@ -38,8 +39,19 @@ def read_secret_value(secret_client, secret_id):
     delsecret = "kubectl -n default delete secret ${secret_name}"
     os.system(delsecret)
 
-    crtsecret = ("kubectl create secret docker-registry ${secret_name} -n default --docker-server=${region_registry} --docker-username=${tenancy_namespace}/${username} --docker-email=${email_address} --docker-password=%s" % secret_content)
-
+    create_namespace = f"""
```
karthicgit, line 42: if namespace is not given it will try to create default . please add a validation to check namespace exists

modules/oke/scripts/secret.py
```diff
@@ -38,8 +39,19 @@ def read_secret_value(secret_client, secret_id):
...
+    cat <<EOF | kubectl apply -f -
+    apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: {namespace}
+      labels:
```
karthicgit, line 48: may i know why these labels are added.

hyder (approved):

PR #324, opened 2021-08-19
secret nullresource trigger will run always

karthicgit:
This PR is to change python to bash script to create secret and the trigger will make the null resource for secret to run every time.


modules/oke/scripts/secret.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Copyright 2017, 2020, Oracle Corporation and/or affiliates.  All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
```
denismakogon, line 4: add
```shell
set -xe
```

hyder:
Please redo from 4.x

PR #334, opened 2021-08-25
allow access to worker nodes from operator host

hyder:


karthicgit (approved):
This will allow ssh access to worker via bastion host and bastion service via operator.if bastion_enabled is set to false still the ingress rule will be created for bastion subnet. in future if bastion_enabled set to true it can be used.

hyder:
> This will allow ssh access to worker via bastion host and bastion service via operator.if bastion_enabled is set to false still the ingress rule will be created for bastion subnet. in future if bastion_enabled set to true it can be used.

The control variables have been renamed btw. It's now `create_bastion_host`

PR #338, opened 2021-08-25
updated quickstart docs so users can start testing

hyder:

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -3,11 +3,13 @@
 
 # Provider parameters
 variable "api_fingerprint" {
+  default     = ""
   description = "Fingerprint of oci api private key."
   type        = string
 }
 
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -3,11 +3,13 @@
 
 # Provider parameters
 variable "api_fingerprint" {
+  default     = ""
   description = "Fingerprint of oci api private key."
   type        = string
 }
 
 variable "api_private_key_path" {
```
karthicgit, line 11: can we add one more variable for private_key in addition to private key path that will be useful in CI/CD.

karthicgit: Also lets use the same name as in provider instead of api_fingerprint etc..

hyder: ok let's do that separately

PR #345, opened 2021-08-25
Changes to support SSH and OCI Keys as text instead of file path

agregory999:
This is how we refer to them in TF Cloud.  We use [sensitive values](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables) to do this.  I marked the variables as such, and when used in a TF Cloud environment, it passes them in.


hyder:
@agregory999 I'm closing this one. We'll work on #351

PR #346, opened 2021-08-27
added support for nsgs for the cluster endpoint

hyder:


variables.tf
```diff
@@ -385,6 +385,12 @@ variable "control_plane_access_source" {
   type        = list(any)
 }
 
+variable "control_plane_nsgs" {
+  default     = []
+  description = "A list of the network security groups (NSGs) ids to apply to the cluster endpoint."
+  type        = list(any)
```
karthicgit, line 391: any reason to make it as list(any) instead of list(string)

hyder: I think that was applied by the linting tool. I'll make it explicitly string

karthicgit (approved):

PR #350, opened 2021-08-30
fix: added 1 extra rule for control plane api 

hyder:
to allow operator to execute kubectl commands


karthicgit (commented):
We need to add one more egress rule for cp subnet for port 6443 for the bastion service to work .

karthicgit (approved):

PR #351, opened 2021-08-30
Changes to support SSH and API Key as file

agregory999:
Hopefully cleaner this time.


main.tf
```diff
@@ -128,7 +128,7 @@ module "policies" {
 
   # ssh keys
   ssh_private_key_path = var.ssh_private_key_path
-  ssh_public_key_path  = var.ssh_public_key_path
+  ssh_private_key = var.ssh_private_key
```
hyder, line 131: Please format the code so they are all aligned

hyder (changes requested):
1. Please remove the .outputs.tf.swp
2. I think you might have missed 1 condition for the private key. See the example on the nodepools https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/4.x/modules/oke/nodepools.tf#L42
3. Since you are adding a couple of new variables, please also update the terraform.tfvars.example 
4. For all new variables, please also update the terraformoptions.adoc

PR #374, opened 2021-09-14
secret changes and added variable validation

karthicgit:
secret.py is removed and secret.sh is created .This change will help in picking the new version created in OCI secrets


hyder (approved):
approved

PR #393, opened 2021-09-24
Make worker subnet security list port definitions more tight

denismakogon:
Closes: #389

### DIFF
--- a/modules/network/locals.tf
+++ b/modules/network/locals.tf
@@ -246,15 +246,6 @@ locals {
       stateless   = false
     },
     {
-      # exceptional rule!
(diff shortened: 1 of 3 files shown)


modules/network/locals.tf
```diff
@@ -156,27 +156,262 @@ locals {
     }
   ]
 
-  workers_ingress = [
+  permitted_ports_for_ingress = [
     {
```
hyder, line 160: Why are you opening these ports? They are not needed

hyder:
Please work only on 1 issue per PR. I'm closing this as this is a critical part of the the project and we need to make sure we fully understand the changes.

Issue #402, opened 2021-10-05
[Feature] Bastion service ID in module outputs


denismakogon:
Please provide bastion service instance ID in module outputs for further session provisioning.

PR #403, opened 2021-10-05
Implement bastion service ID in module output

denismakogon:
Closes: #402


outputs.tf
```diff
@@ -60,3 +60,7 @@ output "kubeconfig" {
   description = "convenient command to set KUBECONFIG environment variable before running kubectl locally"
   value       = "export KUBECONFIG=generated/kubeconfig"
 }
+
+output "bastion_service_id" {
```
hyder, line 64: This should be an instance I think

PR #407, opened 2021-10-05
Fixes: #406 adding egress rule for control plane subnet.

### LINKED ISSUE #406: When setting up a separate OCI bastion service I am unable to connect to the private cluster due to missing egress rule.
### Bug description and details

The control plane subnet security_list does not allow EGRESS within the subnet (which is where the bastion service is).  This blocks access to the bastion service.  I have cluster_access_source set to the entire VCN but this only sets the ingress rule and not the egress.  When I set the egress rule I am able to connect to the kubernetes API via the bastion service session.

```
Terraform v1.0.8
terraform_oci_oke version = "v3.3.0"
registry.terraform.io/hashicorp/oci version = "4.45.0"
```

### Affected Resource(s)

oke_network

### Expected Behavior

I'm using the new bastion service and provisioning it outside this module (since v3.3.0 does not have this feature yet).  I create the bastion service using the module https://registry.terraform.io/providers/hashicorp/oci/latest/docs/resources/bastion_bastion

Bastion service should be able to connect to the control plane subnet and allow access.  ( See: https://www.ateam-oracle.com/post/using-oci-bastion-service-to-manage-private-oke-kubernetes-clusters )

### Actual Behavior

Bastion times out the client connecting to the kubernetes API.

### Steps to Reproduce

Create a new cluster.  Create a bastion service that connects to the control plane subnet.  Attempt to connect to kubernetes following the above guide.

jeremyd:
To allow connectivity within the subnet from the bastion service (fixes #406 )

### DIFF
--- a/modules/okenetwork/locals.tf
+++ b/modules/okenetwork/locals.tf
@@ -38,6 +38,14 @@ locals {
   # control plane
   cp_egress = [
     {
+      description      = "Allow Kubernetes control plane to communicate with other services on this subnet (bastion service)",
+      destination      = local.cp_subnet,
+      destination_type = "CIDR_BLOCK",
+      protocol         = local.tcp_protocol,
+      port             = -1,
+      stateless        = false
+    },
+    {
       description      = "Allow Kubernetes control plane to communicate with OKE",
       destination      = local.osn,
       destination_type = "SERVICE_CIDR_BLOCK",


jeremyd:
Alternatively we could use the variable cluster_access_source for the egress rule. (instead of cp_subnet).  cluster_access_source must be set for bastion connectivity to set the ingress rule (in addition to this egress rule).

modules/okenetwork/locals.tf
```diff
@@ -37,6 +37,14 @@ locals {
 
   # control plane
   cp_egress = [
+    {
+      description      = "Allow Kubernetes control plane to communicate with other services on this subnet (bastion service)",
+      destination      = local.cp_subnet,
+      destination_type = "CIDR_BLOCK",
+      protocol         = local.tcp_protocol,
+      port             = -1,
```
hyder, line 45: This should be port 6443

hyder:
hi @jeremyd,

Thanks for contributing this PR. Before we can accept and merge, can I trouble you to please sign the Oracle Contributor Agreement: https://oca.opensource.oracle.com/?ojr=home

We will then be able to accept and merge your PR.

Thanks again

modules/okenetwork/locals.tf
jeremyd: I tried it with 6443 and it did not work.  I believe this is because it is an egress rule and the traffic is not seen as 'coming from' 6443.

jeremyd:
I signed the OCA.  Thanks!

hyder:
Hi @jeremyd , 

Please advise if this is still a problem.

PR #426, opened 2021-11-09
Added FSS subnet along with security rules between Worker subnet.

ramesumi:
Created network and security list for hosting OCI FSS. Updated worker subnet with an additional security list for FSS mount target network. 

Signed-off-by: Ramesh Ramakrishnan <ramesh.rr.ramakrishnan@oracle.com>

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -70,6 +70,8 @@ module "network" {
   # waf integration
   waf_enabled = var.waf_enabled
 
+  # fss 
+  fss_enabled = var.fss_enabled
 }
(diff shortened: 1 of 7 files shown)


hyder (changes requested):
Looking at your module structure, I think you're still using version 3.x. You need to get an update from main first and make your changes accordingly. Note that for most purposes, we have switched to using NSGs

hyder:
I'll close this PR since you submitted another one.

PR #427, opened 2021-11-11
FSS Provisioning

### LINKED ISSUE #293: Add an FSS subnet with security lists
### Community Note

### Description

Frequently, users of OKE need to add FSS (File System service) for storage purposes. It would be useful if a subnet along with the necessary security lists/nsgs can be added so the network infrastructure is ready to include FSS.

```hcl
```

ramesumi:
Closes #293 
Introduced FSS subnet, security groups and rules for allowing traffic.

Signed-off-by: ramesh ramakrishnan <ramesh.rr.ramakrishnan@oracle.com>


variables.tf
```diff
@@ -770,6 +771,11 @@ variable "service_account_cluster_role_binding" {
   type        = string
 }
 
+# fss mount point network provisioning
+variable "enable_fss" {
+  type = bool
```
hyder, line 776: Please add description and default

karthicgit:
@hyder May i know the benefit of having this fss module inside OKE module instead of a separate module? if we have a separate module for fss that can be beneficial for other projects and also in future if oke brings storage class for auto provision of fss ,oke module  does not need this.

hyder:
@karthicgit those are very good points. But we still need to add the NSG rules for the workers, so I would say at the very least, we need to:

1. conditionally add the subnet 
2. conditionally add the rules to the workers NSG

The above 2 should go into OKE's network module whereas the FSS itself and mount targets should go into a separate sub-module. If it's useful, we can spin it off in its own repo/module. How does that sound?

karthicgit:
@ramesumi  Could you please update the doc for the fss changes.

PR #432, opened 2021-11-29
feat: enhancements to token_helper for kubectl. Closes #429

### LINKED ISSUE #429: Enhancements to token_helper
### Community Note

### Description

With the current design, there are 2 limitations to the token_helper:

1. the token_helper's cluster-id and region parameters are hard-coded
2. the TOKEN_FILE is always ~/.kube/TOKEN

In case of users wanting to use multiple clusters, this becomes a problem. We need to enhance the token_helper so that it always read the cluster-id and region parameters from the current context as well as create the TOKEN_FILE based on the context.

Similarly, when creating/checking for the TOKEN file, a different file needs to be created by cluster.

This would facilitate the use of tools like kubectx and kubens more easily.

```hcl
```

hyder:
Closes #429. Removed hard-coded region and cluster id parameters so the operator can be used to
manage more than 1 OKE cluster. The presence of the token improves the performance of issuing 
kubectl commands to clusters in different regions considerably.

### DIFF
--- a/modules/extensions/scripts/kubeconfig_set_credentials.template.sh
+++ b/modules/extensions/scripts/kubeconfig_set_credentials.template.sh
@@ -2,7 +2,7 @@
 # Copyright 2021 Oracle Corporation and/or affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
-kubectl config set-credentials "user-${cluster-id-11}" --exec-command="./token_helper.sh" \
+kubectl config set-credentials "user-${cluster-id-11}" --exec-command="/usr/local/bin/token_helper.sh" \
   --exec-arg="ce" \
   --exec-arg="cluster" \
   --exec-arg="generate-token" \
--- a/modules/extensions/scripts/kubeconfig_set_credentials.template.sh
+++ b/modules/extensions/scripts/kubeconfig_set_credentials.template.sh
@@ -2,7 +2,7 @@
 # Copyright 2021 Oracle Corporation and/or affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
-kubectl config set-credentials "user-${cluster-id-11}" --exec-command="/usr/local/bin/token_helper.sh" \
+kubectl config set-credentials "user-${cluster-id-11}" --exec-command="$HOME/bin/token_helper.sh" \
   --exec-arg="ce" \
   --exec-arg="cluster" \
   --exec-arg="generate-token" \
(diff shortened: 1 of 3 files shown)


modules/extensions/scripts/token_helper.template.sh
```diff
@@ -2,11 +2,14 @@
 # Copyright 2021 Oracle Corporation and/or affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
-TOKEN_FILE=~/.kube/TOKEN
+CLUSTER=$5
+REGION=$7
+
+TOKEN_FILE=~/.kube/TOKEN-$CLUSTER
```
Djelibeybi, line 8: I recommend using `$HOME` instead of `~` as the latter can break during shell expansion, most obviously under `zsh` when placed inside double quotes (which folks have a tendency to do.

modules/extensions/scripts/kubeconfig_set_credentials.template.sh
```diff
@@ -2,7 +2,7 @@
 # Copyright 2021 Oracle Corporation and/or affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
-kubectl config set-credentials "user-${cluster-id-11}" --exec-command="./token_helper.sh" \
+kubectl config set-credentials "user-${cluster-id-11}" --exec-command="/usr/local/bin/token_helper.sh" \
```
Djelibeybi, line 5: What process is installing the `token_helper.sh` script?

Djelibeybi (changes requested):

modules/extensions/scripts/kubeconfig_set_credentials.template.sh
hyder: Terraform is doing that.

modules/extensions/scripts/token_helper.template.sh
hyder: Done

modules/extensions/scripts/kubeconfig_set_credentials.template.sh
Djelibeybi: Does Terraform have a preferred location for helper scripts/binaries that's more local to the module? Having this in `/usr/local/bin` makes it available to _everyone_, which may be unexpected. It's possibly better placed somewhere within the module hierarchy and sourced using an absolute path.

hyder: this will be invoked mainly by the opc user. so we could perhaps move to to /home/opc/bin?

Djelibeybi: Either that or `/home/opc/.local/bin` perhaps.

hyder: moved to $HOME/bin for opc user

Djelibeybi (approved):

Djelibeybi (approved):

Issue #464, opened 2022-02-03
VCN internet gateway creation can be optional


KSN2510:
Internet gateway is required when either of public subnet or bastion host is being created. But when there is no need for those 2 we need not create internet gateway and can simply disable it

hyder:
closed in #468

PR #468, opened 2022-02-04
feat: Disable internet gateway when lbs are internal and bastion host is disabled

KSN2510:
feat: Disable internet gateway when lbs are internal and bastion host is disabled

Signed-off-by: Nikhil Kota <srinivasa.nikhil.kota@oracle.com>


KSN2510:
closes #464

hyder (approved):
Nice one. Approved.

PR #472, opened 2022-02-10
feat: Added support for adding boot/block volume and in-transit encryption for Operator 

KSN2510:
Added following things
1) Updated the operator to latest version
2) Added support for adding boot/block volume and in-transit encryption for Operator
Signed-off-by: Nikhil Kota <srinivasa.nikhil.kota@oracle.com>

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -250,14 +250,26 @@ The Kubernetes Metrics Server parameter controls the installation of {uri-metric
 
 The KMS integration parameters control whether {uri-oci-kms}[OCI Key Management Service] will be used for encrypting Kubernetes secrets and boot volumes/block volumes. Additionally, the bastion and operator hosts must be enabled as well as instance_principal on the operator.
 
+Bastion Variables
+
 ----
 create_bastion_host = true
-create_operator = true
-enable_operator_instance_principal = true
+----
+
+OKE Variables:-
+
+----
 use_encryption = true
 kms_key_id = <kms_key_id>
 enable_pv_encryption_in_transit = false
 node_pool_volume_kms_key_id = <node_pool_volume_kms_key_id>
+----
+
+Operator Variables
+
+----
+create_operator = true
+enable_operator_instance_principal = true
 enable_operator_pv_encryption_in_transit = false
 operator_volume_kms_id = <operator_volume_kms_id>
 ----
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -258,6 +258,8 @@ use_encryption = true
 kms_key_id = <kms_key_id>
 enable_pv_encryption_in_transit = false
 node_pool_volume_kms_key_id = <node_pool_volume_kms_key_id>
+enable_operator_pv_encryption_in_transit = false
+operator_boot_volume_encryption_key_id = <operator_boot_volume_encryption_key_id>
 ----
 
 OKE also supports enforcing the use of signed images. You can enforce the use of signed image using the following parameters:
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -259,7 +259,7 @@ kms_key_id = <kms_key_id>
 enable_pv_encryption_in_transit = false
 node_pool_volume_kms_key_id = <node_pool_volume_kms_key_id>
 enable_operator_pv_encryption_in_transit = false
-operator_boot_volume_encryption_key_id = <operator_boot_volume_encryption_key_id>
+operator_volume_kms_id = <operator_volume_kms_id>
 ----
 
 OKE also supports enforcing the use of signed images. You can enforce the use of signed image using the following parameters:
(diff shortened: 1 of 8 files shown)


karthicgit (approved):

docs/configuration.adoc
```diff
@@ -258,6 +258,8 @@ use_encryption = true
 kms_key_id = <kms_key_id>
 enable_pv_encryption_in_transit = false
 node_pool_volume_kms_key_id = <node_pool_volume_kms_key_id>
+enable_operator_pv_encryption_in_transit = false
+operator_volume_kms_id = <operator_volume_kms_id>
```
hyder, line 262: You need to add subsections here so you can distinguish between bastion, operator and oke's respective KMS integration. 

Right now, this section tells the user what parameters they need in order to do this integration for OKE **only**. Since you're adding similar capability for bastion and operator, add subsections for them

docs/instructions.adoc
```diff
@@ -78,10 +78,18 @@ If you wish to use {uri-oci-kms}[OCI KMS] to encrypt OKE nodepool boot/block vol
...
 * enable_pv_encryption_in_transit must be _true_
 
+If you wish to use {uri-oci-kms}[OCI KMS] to encrypt operator boot/block volume, the following is required:
+
+* operator_volume_kms_id must be provided
+
+If you wish to encrypt data in transit between the instance, the boot volume, and the block volumes in operator.
+
+* enable_operator_pv_encryption_in_transit must be _true_
+
```
hyder, line 92: As above, add subsections so the user know what is being encrypted. Don't rely on the variable name.

docs/terraformoptions.adoc
```diff
@@ -407,6 +407,16 @@ EOT
 | RUNNING/STOPPED
 | RUNNING
 
+|enable_operator_pv_encryption_in_transit
```
hyder, line 410: Please enclose the name of a variable between backticks so it renders like `my_variable`

hyder: Please group variables together, within each group sort them in alphabetical order

docs/terraformoptions.adoc
```diff
@@ -407,6 +407,16 @@ EOT
...
+|Whether to encrypt data in transit between the instance, the boot volume, and the block volumes in Operator.
+|true/false
+|false
+
+|operator_volume_kms_id
```
hyder, line 415: If there's a dependency on a variable defined to be of a certain value, please mention it here.

modules/oke/locals.tf
```diff
@@ -18,7 +18,7 @@ locals {
   policy_statement = (var.use_encryption == true) ? "Allow dynamic-group ${oci_identity_dynamic_group.oke_kms_cluster[0].name} to use keys in compartment id ${var.compartment_id} where target.key.id = '${var.kms_key_id}'" : ""
 
   # policy to allow block volumes inside oke to use kms
-  oke_volume_kms_policy_statements = [
+  oke_volume_kms_policy_statements = (var.node_pool_volume_kms_key_id == "") ? []: [
```
hyder, line 21: You are submitting a PR for operator but also changing a functionality in OKE. In future PRs, please change only that for which you are submitting a PR.

terraform.tfvars.example
```diff
@@ -104,6 +104,10 @@ operator_state    = "RUNNING"
 operator_timezone = "Australia/Sydney"
 upgrade_operator  = false
 
+### operator volume kms integration
+enable_operator_pv_encryption_in_transit = false
```
hyder, line 108: As above, group related variables together, within the group sort in alphabetical order. If the group is big enough to warrant its own, do a subsection see operator notification below. They all still belong to the operator but the set of parameters below are for operator notification only

variables.tf
```diff
@@ -296,6 +296,18 @@ variable "operator_image_id" {
...
+variable "operator_volume_kms_id" {
+  default     = ""
+  description = "The OCID of the OCI KMS key to assign as the master encryption key for the boot volume."
+  type        = string
+}
+
+variable "enable_operator_pv_encryption_in_transit" {
+  default     = false
+  description = "Whether to enable in-transit encryption for the data volume's paravirtualized attachment."
+  type        = bool
```
hyder, line 308: Please add validation

KSN2510: what do you mean by validation here ? enable_operator_pv_encryption_in_transit is independent on other features and it can be either true or false which is already captured by bool type

modules/oke/locals.tf
KSN2510: sure, will keep that in mind

docs/terraformoptions.adoc
KSN2510: there is no dependency here

docs/terraformoptions.adoc
KSN2510: done

variables.tf
hyder: see here for an example: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/variables.tf#L223

hyder (approved):
approved

PR #487, opened 2022-02-23
feat: renamed kms_key_id to cluster_kms_key_id to avoid confusion. 

### LINKED ISSUE #458: Rename kms_key_id to something appropriate
The key name of the cluster kms key is **kms_key_id** it would be better if we change it to something appropriate like **cluster_kms_key_id** since there is a different key for encrypting block volume/boot volume in nodepool with name **node_pool_volume_kms_key_id**, and also we are creating a **kms_key_id** in extensions module but not using that anywhere inside module which can be removed from there

hyder:
Closes #458

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -260,7 +260,7 @@ OKE Variables:-
 
 ----
 use_encryption = true
-kms_key_id = <kms_key_id>
+cluster_kms_key_id = <cluster_kms_key_id>
 use_node_pool_volume_encryption = true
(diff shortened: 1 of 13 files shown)


karthicgit:
@hyder This is already done as part of PR 483

hyder:
This is for etcd, not the nodepool. Etcd encryption's current variable is still kms_key_id

karthicgit:
yes i mean the  cluster etcd. i have made the changes as part of fss PR .

karthicgit (approved):

Issue #488, opened 2022-02-23
Operator creation failing with a 404-NotAuthorizedOrNotFound error


Deepakkoli93:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version and Provider Version
Terraform v1.0.7
provider registry.terraform.io/hashicorp/oci v4.64.0

### Affected Resource(s)

`oci_core_instance`

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. 
# Please remove any sensitive information from configuration files before sharing them. 
module "oci_oke" {
  source                          = "oracle-terraform-modules/oke/oci"
  version                         = "4.1.5"
  compartment_id                  = var.compartment_id
  region                          = var.region
  home_region                     = local.home_region
  tenancy_id                      = var.tenancy_ocid
  ssh_private_key                 = local.private_ssh_key
  ssh_public_key                  = local.public_ssh_key
  vcn_name                        = "My-VCN"
  vcn_dns_label                   = "vcnMyVcn"
  create_bastion_host             = false
  create_operator                 = true
  label_prefix                    = "test"
  providers                       = {
    oci.home = oci.home
  }
  vcn_cidrs                                = local.vcn_cidrs
  control_plane_type                       = "private"
  pods_cidr                                = local.pods_cidr
  services_cidr                            = local.services_cidr
  use_encryption                           = true
  kms_key_id                               = var.existent_encryption_key_id
  enable_pv_encryption_in_transit          = true
  node_pool_volume_kms_key_id              = var.existent_encryption_key_id
  enable_operator_pv_encryption_in_transit = true
  use_node_pool_volume_encryption          = true
  operator_volume_kms_id                   = var.existent_encryption_key_id
  freeform_tags                            = var.freeform_tags
  load_balancers                           = "internal"
  preferred_load_balancer                  = "internal"
  subnets                                  = local.subnets
  node_pools                               = local.node_pools
  create_bastion_service                   = local.create_bastion_service
  bastion_service_name                     = local.bastion_service_name
[...]

PR #489, opened 2022-02-23
WIP: fix: fixed operator 404 issue while using operator pool volume encryp…

Deepakkoli93:
Made sure that operator is created after k8s cluster so that the following required policy is in place before operator is created
`Allow service blockstorage to use keys in compartment id ${var.compartment_id} where target.key.id = '${var.node_pool_volume_kms_key_id}'`

Signed-off-by: Deepak Koli <deepak.koli@oracle.com>

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -127,7 +127,7 @@ module "operator" {
   }
 
   depends_on = [
-    module.vcn
+    module.vcn, module.oke
   ]
 
   count = var.create_operator == true ? 1 : 0


hyder:
As discussed, the `operator` module is independent and it should not depend on the `oke` module. As @KSN2510 is the original contributor to the operator volume encryption feature, can you please discuss with him to find why this is happening?

hyder:
Closes #488

PR #500, opened 2022-03-22
removed null resource for localkubeconfig and helm

karthicgit:
null resource is not needed for local kubeconfig creation since local_file resource will create it if does not exist. Merged helm install commands into kubectl null resource.

### DIFF
--- a/modules/extensions/calico.tf
+++ b/modules/extensions/calico.tf
@@ -14,7 +14,7 @@ resource "null_resource" "install_calico" {
     bastion_private_key = local.ssh_private_key
   }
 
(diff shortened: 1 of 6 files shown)


hyder (changes requested):
1. Since you're installing both kubectl and helm with the same null_resource, please rename it to install_k8s_tools
2. Also, since most of the extensions here are dependent on the bastion and operator running and enabled, can we check for the conditions on the module itself in main instead of each extension we add?

karthicgit:
renamed null_resource.  I can add the condition later in another PR.

hyder (approved):
Approved

Issue #515, opened 2022-04-21
Allow specifying the ADs each node pool is placed in


12345ieee:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It's currently not possible to choose which ADs a nodepool is placed in.
All the ADs are enabled, via the code:
```hcl
    dynamic "placement_configs" {
      iterator = ad_iterator
      for_each = local.ad_names
      content {
        availability_domain = ad_iterator.value
        subnet_id           = var.cluster_subnets["workers"]
      }
    }
```

### New or Affected Resource(s)

```
oci_containerengine_node_pool
```

### Potential Terraform Configuration

I propose adding a new parameter to the `node_pools` submap to specify the AD list.
An empty parameter would mean every AD, as it's now to not break existing configurations:

```hcl
  node_pools = {
    pool1_name = { shape = "shape", ocpus = 2, memory = 32, node_pool_size = 1, label = { mylabel = "myvalue" }, 
                  placement_ADs = [AD1] }, # pool only on AD1
    pool2_name = { shape = "shape", ocpus = 2, memory = 32, node_pool_size = 1, label = { mylabel = "myvalue" }, 
                  placement_ADs = [AD1, AD2] }, # pool on AD1,2
    pool3_name = { shape = "shape", ocpus = 2, memory = 32, node_pool_size = 1, label = { mylabel = "myvalue" } },
                   # pool on all ADs
  }
```

the placement code would then become something like:
```hcl
    dynamic "placement_configs" {
      iterator = ad_iterator
      for_each = lookup(each.value, "placement_ADs", "") != "" ? set(lookup(each.value, "placement_ADs", "")) : local.ad_names
      content {
        availability_domain = ad_iterator.value
        subnet_id           = var.cluster_subnets["workers"]
      }
    }
```

hyder:
Hi @12345ieee,

This looks good to me. Would you be willing to send us a PR? 

Just out of curiosity, can you please elaborate on your use case?

12345ieee:
> This looks good to me. Would you be willing to send us a PR?
[...]

PR #520, opened 2022-05-10
Added option to specify ADs in node pools

12345ieee:
I haven't been able to test it yet, as I'm still working on getting a test environment up, but in the meantime the maintainers can tell if there are issues with the documentation or the commit message.

----
The `node_pools` argument gained a new optional key `placement_ADs`.
This key can be popolated by a list, like:
`placement_ADs = ["sQTi:EU-FRANKFURT-1-AD-1"]`
to choose the ADs the pool will deploy nodes to.
Default remains to deploy to all ADs of region.

Fixes: #515

Signed-off-by: Andrea Stacchiotti <andreastacchiotti@gmail.com>

### DIFF
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -739,14 +739,9 @@ Refer to {uri-topology}[topology] for more thorough examples.
 node_pools = {
 np1 = { shape = "VM.Standard.E4.Flex", ocpus = 1, memory = 16, node_pool_size = 1, boot_volume_size = 150, label = { app = "frontend", pool = "np1" } }
   np2 = {shape="VM.Standard.E2.2",node_pool_size=2,boot_volume_size=150,label={app="application",name="test"}}
-  np3 = {shape="VM.Standard.E2.2",node_pool_size=1}
-}
-|
-node_pools = {
-  np1 = {shape="VM.Standard.E3.Flex",ocpus=2,node_pool_size=2,boot_volume_size=150}
-  np2 = {shape="VM.Standard.E2.2",node_pool_size=2,boot_volume_size=150,label={app="application",name="test"}}
-  np3 = {shape="VM.Standard.E2.2",node_pool_size=1}
-}
+  np3 = {shape="VM.Standard.E2.2",node_pool_size=1,placement_ads=[1]} 
+} 
+|{}
 
 |node_pool_image_id
 |The OCID of custom image to use when provisioning worker nodes. When no OCID is specified, the worker nodes will use the node_pool_os and node_pool_os_version to identify an image to provision the worker nodes.
--- a/modules/oke/locals.tf
+++ b/modules/oke/locals.tf
@@ -6,10 +6,11 @@ locals {
   # used by cluster
   lb_subnet = var.preferred_load_balancer == "public" ? "pub_lb" : "int_lb"
 
-  ad_names = [
-    for ad_name in data.oci_identity_availability_domains.ad_list.availability_domains :
-    ad_name.name
-  ]
+  ad_number_to_name = {
+    for ad in data.oci_identity_availability_domains.ad_list.availability_domains :
+    parseint(substr(ad.name, -1, -1), 10) => ad.name
+  }
+  ad_numbers = keys(local.ad_number_to_name)
 
   # dynamic group all oke clusters in a compartment
   dynamic_group_rule_all_clusters = "ALL {resource.type = 'cluster', resource.compartment.id = '${var.compartment_id}'}"
--- a/modules/oke/nodepools.tf
+++ b/modules/oke/nodepools.tf
@@ -21,7 +21,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     # iterating over ADs
     dynamic "placement_configs" {
       iterator = ad_iterator
-      for_each = local.ad_names
+      for_each = [for n in lookup(each.value, "placement_ads", local.ad_numbers):
+                  local.ad_number_to_name[n]]
       content {
         availability_domain = ad_iterator.value
         subnet_id           = var.cluster_subnets["workers"]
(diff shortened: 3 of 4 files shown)


karthicgit:
@12345ieee Lets have a separate variable for this since the nodepool variable already have many keys and in future if choice is given for FD in single AD it can be added in the new variable and other options like capacity reservation etc. Another suggestion would be to use datasource to fetch the AD values instead of finding the value and hardcoding it  . For example: placementconfig = {np1 = [AD-1]} and filter the datasource to get the value . May i know what is the use-case for using single AD placement config.

12345ieee:
@karthicgit thanks for the comments:

I disagree with having a different variable for ADs, as one could want to deploy node pool 1 on AD1, node pool 2 on AD2 and node pool 3 to all ADs. The ADs need to be a property of each pool.
The pool subobject is indeed getting pretty long, but that's better solved by recommending people to version a pretty printed json and then using `jsondecode` to fill the nested map.

My intention was for the module user to hold the AD datasource, not having it in the module, but didn't make it explicit in the docs. You'd normally use the module as such:
```tf
data "oci_identity_availability_domain" "AD1" {
  compartment_id = data.oci_identity_compartment.root.id
  ad_number      = "1"
}

module "oke" {
...
  {..., placement_ADs = [data.oci_identity_availability_domain.AD1.name], ... }
}
```
as I imagine any user doing nontrivial work on oci has the AD datasources laying around already.

My rationale for this feature is here: https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/515#issuecomment-1114369120

karthicgit:
@12345ieee  while using the same nodepool key it will be mapped . With a new variable you will map with the same np1 or other value used for the key in nodepool variable .

12345ieee:
I see your point now @karthicgit .
If we're splitting the map in 2 it may make sense to move the content of the whole `node_config_details` block into the new map, resulting in something like:

```tf
node_pools = {
  np1 = {shape="VM.Standard.E3.Flex",ocpus=2,boot_volume_size=150}
}

node_pools_config_details = {
  np1 = {node_pool_size=2,placement_ADs=["sQTi:EU-FRANKFURT-1-AD-1"]}
}
```
this also gives us a path forward into adding more options, by splitting each nodepool configuration block into its argument map.

If you like this option better I can implement it.

12345ieee:
So, what would be the preferred solution @karthicgit ?

karthicgit:
@12345ieee i would prefer not changing the existing keys and add only the new ones in new variable. 
@hyder  could you please share your inputs.

12345ieee:
I think for now the easiest way is to proceed with the simple change, deeper restructuring can be done on top.

karthicgit:
@12345ieee As discussed please create a new variable and also don't hardcode the AD values .we have to fetch it from input like AD1,AD2 etc..  Fault domain support in placement config is also available now . https://docs.oracle.com/en-us/iaas/releasenotes/changes/d13cef7f-42fa-4313-be4d-b6ff117f7997/ .

karthicgit:
@12345ieee As discussed please create a new variable and also don't hardcode the AD values .we have to fetch it from input like AD1,AD2 etc..  Fault domain support in placement config is also available now . https://docs.oracle.com/en-us/iaas/releasenotes/changes/d13cef7f-42fa-4313-be4d-b6ff117f7997

hyder:
hey @12345ieee,

Where are you with this? Another user just reached out to me about the possibility of pinning a node pool to an AD.

I note that it's been quite a while so your branch is very likely well out of date considering the recent changes we did to nodepools. Let us know if you need any help. I would like to move this forward.

12345ieee:
Hi @hyder ,

the code has been ready for a while, but I haven't been able to test it in action.
I have also not implemented the additional variable that was requested, but I've added the key to the np map.

I've just pushed what I have, the intent should be clear and you should be able to build on it if needed, I hope to be able to test it if still needed in the following weeks.

docs/terraformoptions.adoc
```diff
@@ -739,14 +739,9 @@ Refer to {uri-topology}[topology] for more thorough examples.
...
   np2 = {shape="VM.Standard.E2.2",node_pool_size=2,boot_volume_size=150,label={app="application",name="test"}}
-  np3 = {shape="VM.Standard.E2.2",node_pool_size=1}
-}
-|
-node_pools = {
-  np1 = {shape="VM.Standard.E3.Flex",ocpus=2,node_pool_size=2,boot_volume_size=150}
-  np2 = {shape="VM.Standard.E2.2",node_pool_size=2,boot_volume_size=150,label={app="application",name="test"}}
-  np3 = {shape="VM.Standard.E2.2",node_pool_size=1}
-}
+  np3 = {shape="VM.Standard.E2.2",node_pool_size=1,placement_ADs=[1]}
```
hyder, line 742: Please use snake_case like the other parameters

hyder (changes requested):
Please rebase and then push to your fork again

hyder:
@12345ieee: We also need you to sign the OCA.

docs/terraformoptions.adoc
12345ieee: This is snake_case in my mind, would you prefer `placement_ads`?

hyder: yes please

modules/oke/locals.tf
```diff
@@ -6,10 +6,11 @@ locals {
   # used by cluster
   lb_subnet = var.preferred_load_balancer == "public" ? "pub_lb" : "int_lb"
 
-  ad_names = [
-    for ad_name in data.oci_identity_availability_domains.ad_list.availability_domains :
-    ad_name.name
-  ]
+  ad_number_to_name = {
+    for ad in data.oci_identity_availability_domains.ad_list.availability_domains :
+    ad.ad_number => ad.name
```
hyder, line 11: There's no attribute `ad_number`. See https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/identity_availability_domains

modules/oke/locals.tf
```diff
@@ -6,10 +6,11 @@ locals {
...
+  }
+  ad_numbers = keys(ad_number_to_name)
```
hyder, line 13: This should be referenced as `local.ad_number_to_name)` once you've figured out how to resolve the above problem

modules/oke/nodepools.tf
```diff
@@ -21,7 +21,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     # iterating over ADs
     dynamic "placement_configs" {
       iterator = ad_iterator
-      for_each = local.ad_names
+      for_each = [for n in lookup(each.value, "placement_ads", local.ad_numbers):
+                  ad_number_to_name[n]]
```
hyder, line 25: This should be referenced as local.ad_number_to_name) once you've figured out how to resolve the above problem

hyder (changes requested):
Please check and test the code again. There are some errors.

12345ieee:
Yeah, this is what happens when I push untested code.

I'm not super happy about what I did for ad numbers, but that's what the API offers.

hyder (approved):

hyder:
Thanks @12345ieee for your patience.

PR #548, opened 2022-07-28
feat: added kubectx and kubens

### LINKED ISSUE #547: Add kubectx and kubens as tools on operator
### Community Note

### Description
kubectx and kubens are 2 very useful Kubernetes utilities that help Kubernetes admins change context and namespaces quickly and easily without having to remember the long version, especially in environments where many clusters are deployed. We propose to add them as well as aliases for them to facilitate their use.

```hcl
```

hyder:
Closes #547


modules/extensions/k8stools.tf
```diff
@@ -23,14 +23,23 @@ resource "null_resource" "install_k8stools_on_operator" {
     content     = local.install_helm_template
     destination = "/home/opc/install_helm.sh"
   }
+
+  provisioner "file" {
+    content     = local.install_kubectx_template
+    destination = "/home/opc/install_kubectx.sh"
```
Djelibeybi, line 29: You're hard-coding a path here, but below you use `$HOME/install_kubectx.sh` and I see you have a hard-coded path for `install_helm.sh` above. I would recommend using `$HOME` throughout so that you're not tied to the specific username of the generic user.

modules/extensions/k8stools.tf
```diff
@@ -23,14 +23,23 @@ resource "null_resource" "install_k8stools_on_operator" {
...
+  }
+
   provisioner "remote-exec" {
     inline = [
       "chmod +x $HOME/install_kubectl.sh",
       "$HOME/install_kubectl.sh",
       "rm -f $HOME/install_kubectl.sh",
```
Djelibeybi, line 36: You probably should check that `$HOME` contains a value that is meaningful otherwise the `rm -f` may run against `/install_kubectl.sh`.

modules/extensions/k8stools.tf
```diff
@@ -23,14 +23,23 @@ resource "null_resource" "install_k8stools_on_operator" {
...
       "chmod +x $HOME/install_helm.sh",
       "$HOME/install_helm.sh",
-      "rm -f $HOME/install_helm.sh"
+      "rm -f $HOME/install_helm.sh",
+      "chmod +x $HOME/install_kubectx.sh",
+      "$HOME/install_kubectx.sh",
```
Djelibeybi, line 41: Instead of making the file executable, why not just run it via `bash $HOME/install_kubectx.sh`?

modules/extensions/scripts/install_kubectx.template.sh
```diff
@@ -0,0 +1,18 @@
...
+
+curl -sSfL https://github.com/ahmetb/kubectx/releases/download/v0.9.4/kubectx_v0.9.4_linux_x86_64.tar.gz -o kubectx_v0.9.4_linux_x86_64.tar.gz
+curl -sSfL https://github.com/ahmetb/kubectx/releases/download/v0.9.4/kubens_v0.9.4_linux_x86_64.tar.gz -o kubens_v0.9.4_linux_x86_64.tar.gz
+
+tar zxf  kubectx_v0.9.4_linux_x86_64.tar.gz
+tar zxf kubens_v0.9.4_linux_x86_64.tar.gz
+rm -f  kubectx_v0.9.4_linux_x86_64.tar.gz
+rm -f kubens_v0.9.4_linux_x86_64.tar.gz
+rm -f LICENSE
+
```
Djelibeybi, line 13: I recommend parameterisaing the version so you don't have to edit all the commands every time it updates, just the version variable. Also, the repo supports multiple operating systems and architectures, so let's plan for the future and enable that too:

```bash
KUBECTX_VERSION="0.9.4"
KUBENS_VERSION="0.9.4"
ARCH=$(uname -m)
```
Then, you could combine all of this into two commands:

```bash
curl -sSfL https://github.com/ahmetb/kubectx/releases/download/v${KUBECTX_VERSION}/kubectx_v${KUBECTX_VERSION}_linux_${ARCH}.tar.gz | tar zxvf - -C /usr/local --xform='s|^|kubectx/|S'
curl -sSfL https://github.com/ahmetb/kubectx/releases/download/v${KUBENS_VERSION}/kubens_v${KUBENS_VERSION}_linux_${ARCH}.tar.gz | tar zxvf - -C /usr/local --xform='s|^|kubens/|S'
```

This will create `/usr/local/kubectx/{kubectx,LICENSE}` and `/usr/local/kubens/{kubens,LICENSE}` and retains the `LICENSE` for each tool, which is mandatory. You can link them to `/usr/local/bin` and alias as needed.

modules/extensions/scripts/install_kubectx.template.sh
```diff
@@ -0,0 +1,18 @@
...
+sudo mv kubectx /usr/local/bin
+sudo mv kubens /usr/local/bin
```
Djelibeybi, line 15: Instead of moving, create symbolic links to `/usr/bin`:

```bash
ln -s /usr/bin/kubectx /usr/local/kubectx/kubectx
ln -s /usr/bin/kubens /usr/local/kubens/kubens
```

Djelibeybi (changes requested):

modules/extensions/k8stools.tf
hyder: $HOME gets rendered as a string and the file is not written.

modules/extensions/k8stools.tf
hyder: done

modules/extensions/scripts/install_kubectx.template.sh
hyder: I had this before but Terraform then expects me to pass the value. Ditto for ARCH. I've changed this somewhat. Please check again.

modules/extensions/scripts/install_kubectx.template.sh
hyder: Ok but I think it should be 

```
sudo ln -s /usr/local/kubectx/kubectx /usr/bin/kubectx

sudo ln -s /usr/local/kubens/kubens /usr/bin/kubens 
```

modules/extensions/k8stools.tf
```diff
@@ -23,14 +23,23 @@ resource "null_resource" "install_k8stools_on_operator" {
...
+  provisioner "file" {
+    content     = local.install_kubectx_template
+    destination = "/home/opc/install_kubectx.sh"
+  }
+
   provisioner "remote-exec" {
     inline = [
-      "chmod +x $HOME/install_kubectl.sh",
-      "$HOME/install_kubectl.sh",
+      # "chmod +x $HOME/install_kubectl.sh",
```
Djelibeybi, line 34: It's considered bad practice leaving commented out code in the source. If you need to know what is was, that's what git is for. 😁

modules/extensions/scripts/install_kubectx.template.sh
Djelibeybi: You're right. I always get symlinks the wrong way around.

modules/extensions/k8stools.tf
hyder: Fixed and did the same for other scripts.

modules/extensions/k8stools.tf
hyder: for example?

modules/extensions/activeworker.tf
```diff
@@ -29,7 +29,6 @@ resource "null_resource" "check_worker_active" {
     inline = [
       "chmod +x $HOME/check_active_worker.sh",
       "$HOME/check_active_worker.sh"
-      //"rm -f $HOME/check_active_worker.sh"
```
Djelibeybi, line 32: Remove this commented out line.

modules/extensions/k8stools.tf
Djelibeybi: Something like this would test that the file exists before attempting to either run it or delete it.

```bash
if [ -f "$HOME/install_kubectl.sh" ]; then bash $HOME/install_kubectl.sh; fi
```

In the case of your `rm` command, something like:

```bash
if [ -f "$HOME/install_kubectl.sh" ]; then rm -f $HOME/install_kubectl.sh; fi
```

modules/extensions/k8stools.tf
Djelibeybi: I'm not seeing the fix in this PR?

modules/extensions/activeworker.tf
```diff
@@ -27,8 +27,7 @@ resource "null_resource" "check_worker_active" {
 
   provisioner "remote-exec" {
     inline = [
-      "chmod +x $HOME/check_active_worker.sh",
-      "$HOME/check_active_worker.sh"
+      "if [ -f \"$HOME/check_active_worker.sh\" ]; then bash $HOME/check_active_worker.sh; rm -f \"$HOME/check_active_worker.sh\";fi",
```
Djelibeybi, line 30:
```suggestion
      'if [ -f "$HOME/check_active_worker.sh" ]; then bash "$HOME/check_active_worker.sh" && rm -f "$HOME/check_active_worker.sh"; fi',
```

Though, I'm not sure if HCL allows for single quoted strings in lists to preserve double quotes. Hopefully it does. If so, this pattern can be repeated for the rest, as we really do want to wrap the path to the script, if possible.

Djelibeybi (approved):

PR #562, opened 2022-08-29
feat: added tagging using defined tags #560

### LINKED ISSUE #560: Support defined tags in addition to freeform
### Community Note

### Description

The module supports adding freeform tags to all of the resources created by the module, but has no way to add defined tags.
Defined tags are used by the bean counters, it'd be very helpful to be able to set them in the module.

All the module children.

An exact clone of the freeform tags map:

```hcl
defined_tags = {
  # vcn, bastion and operator freeform_tags are required
  # add more defined_tags in each as desired
  vcn = {
    environment = "dev"
  }
  bastion = {
    access      = "public",
    environment = "dev",
    role        = "bastion",
    security    = "high"
  }
  operator = {
    access      = "restricted",
    environment = "dev",
    role        = "operator",
    security    = "high"
  }
  oke = {
    cluster = {
      environment = "dev"
      role        = "cluster"
    }
    persistent_volume = {
      environment = "dev"
    }    
    service_lb = {
      environment = "dev"
      role        = "load balancer"
    }
    node_pool = {
      environment = "dev"
      role        = "node-pool"
    }
  }
}
```

### References

See #538 .

valireds:
Closes #560 
Signed-off-by: Valentina Rossi <valentina.rossi@sighup.io>

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -35,6 +35,8 @@
 :uri-oracle: https://www.oracle.com
 :uri-prereqs: {uri-docs}/prerequisites.adoc
 :uri-quickstart: {uri-docs}/quickstart.adoc
+:uri-oci-tags-overview: https://docs.oracle.com/en-us/iaas/Content/Tagging/Concepts/taggingoverview.htm
+:uri-oci-tags-management: https://docs.oracle.com/en-us/iaas/Content/Tagging/Tasks/managingtagsandtagnamespaces.htm#workdefined
 
 :uri-terraform: https://www.terraform.io
 :uri-terraform-cidrsubnet-desconstructed: http://blog.itsjustcode.net/blog/2017/11/18/terraform-cidrsubnet-deconstructed/
@@ -286,3 +288,8 @@ image_signing_keys = ["ocid1.key.oc1....", "ocid1.key.oc1...."]
 ----
 {uri-terraform-options}#kms-integration[Reference]
 
+== Configure defined tags
+
+The {uri-oci-tags-overview}#[defined tags] *must* first be manually created before using them as input parameters. 
+If a tag is a list type, you must specify one of the predefined options when applying it. 
+Please refer to the {uri-oci-tags-management}[instructions] for how to assign the required permissions and to manage them.
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -574,6 +574,49 @@ EOT
     }
   }
 }`
+
+|defined_tags
+|Tagging using administrator defined tags. This is specified in the form of a map of maps.
+| e.g.
+`defined_tags = {
+  vcn = {
+    environment = "dev"
+  }
+  bastion = {
+    environment = "dev"
+    role        = "bastion"
+  }
+  operator = {
+    environment = "dev"
+    role        = "operator"
+  }
+  oke = {
+    service_lb  = {
+      environment = "dev"
+      role        = "load balancer"
+    }
+  }
+}`
+|
+`defined_tags = {
+  vcn = {
+    environment = "dev"
+  }
+  bastion = {
+    environment = "dev"
+    role        = "bastion"
+  }
+  operator = {
+    environment = "dev"
+    role        = "operator"
+  }
+  oke = {
+    service_lb  = {
+      environment = "dev"
+      role        = "load balancer"
+    }
+  }
+}`
 |===
 
 == OKE
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -20,6 +20,7 @@
 :uri-oci-kms: https://docs.cloud.oracle.com/iaas/Content/KeyManagement/Concepts/keyoverview.htm
 :uri-oci-loadbalancer-annotations: https://github.com/oracle/oci-cloud-controller-manager/blob/master/docs/load-balancer-annotations.md
 :uri-oci-region: https://docs.cloud.oracle.com/iaas/Content/General/Concepts/regions.htm
+:uri-oci-tags: https://docs.oracle.com/en-us/iaas/Content/Tagging/Tasks/managingtagsandtagnamespaces.htm#workdefined
 :uri-terraform-cidrsubnet: https://www.terraform.io/docs/configuration/functions/cidrsubnet.html
 :uri-terraform-dependencies: {uri-docs}/dependencies.adoc
 :uri-timezones: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
@@ -576,29 +577,28 @@ EOT
 }`
 
 |defined_tags
-|Tagging using administrator defined tags. This is specified in the form of a map of maps.
+|Tagging using administrator defined tags, which *must* already exist. This is specified in the form of a map of maps, using format _"namespace.key" = "value"_ for a specific tag. For further information on tags, please refer to {uri-oci-tags}[the official documentation].
 | e.g.
 `defined_tags = {
   vcn = {
-    environment = "dev"
+    "cn.env" = "dev"
   }
   oke = {
     service_lb  = {
-      environment = "dev"
-      role        = "load balancer"
+      "cn.env" = "dev"
+      "cn.role" = "load balancer"
     }
   }
 }`
 |
 `defined_tags = {
-  vcn = {
-    environment = "dev"
-  }
+  vcn = {}
   oke = {
-    service_lb  = {
-      environment = "dev"
-      role        = "load balancer"
-    }
+    cluster = {}
+    persistent_volume = {}
+    service_lb = {}
+    node_pool = {}
+    node = {}
   }
 }`
 |===
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -582,14 +582,6 @@ EOT
   vcn = {
     environment = "dev"
   }
-  bastion = {
-    environment = "dev"
-    role        = "bastion"
-  }
-  operator = {
-    environment = "dev"
-    role        = "operator"
-  }
   oke = {
     service_lb  = {
       environment = "dev"
@@ -601,14 +593,6 @@ EOT
 `defined_tags = {
   vcn = {
     environment = "dev"
-  }
-  bastion = {
-    environment = "dev"
-    role        = "bastion"
-  }
-  operator = {
-    environment = "dev"
-    role        = "operator"
   }
   oke = {
     service_lb  = {
--- a/main.tf
+++ b/main.tf
@@ -95,7 +95,6 @@ module "bastion" {
 
   # tags
   freeform_tags = var.freeform_tags["bastion"]
-  defined_tags  = var.defined_tags["bastion"]
 
   providers = {
     oci.home = oci.home
@@ -147,7 +146,6 @@ module "operator" {
 
   # tags
   freeform_tags = var.freeform_tags["operator"]
-  defined_tags  = var.defined_tags["operator"]
 
   providers = {
     oci.home = oci.home
--- a/main.tf
+++ b/main.tf
@@ -19,8 +19,9 @@ module "vcn" {
   # lpgs
   local_peering_gateways = var.local_peering_gateways
 
-  # freeform_tags
+  # tags
   freeform_tags = var.freeform_tags["vcn"]
+  defined_tags  = var.defined_tags["vcn"]
 
   # vcn
   vcn_cidrs                    = var.vcn_cidrs
@@ -92,7 +93,9 @@ module "bastion" {
   bastion_notification_protocol = var.bastion_notification_protocol
   bastion_notification_topic    = var.bastion_notification_topic
 
+  # tags
   freeform_tags = var.freeform_tags["bastion"]
+  defined_tags  = var.defined_tags["bastion"]
 
   providers = {
     oci.home = oci.home
@@ -142,7 +145,9 @@ module "operator" {
   operator_notification_protocol = var.operator_notification_protocol
   operator_notification_topic    = var.operator_notification_topic
 
+  # tags
   freeform_tags = var.freeform_tags["operator"]
+  defined_tags  = var.defined_tags["operator"]
 
   providers = {
     oci.home = oci.home
@@ -280,7 +285,9 @@ module "oke" {
   pod_nsgs = concat(module.network.pod_nsg_id)
   worker_nsgs = concat(var.worker_nsgs, [module.network.worker_nsg_id])
 
+  # tags
   freeform_tags = var.freeform_tags["oke"]
+  defined_tags  = var.defined_tags["oke"]
 
   depends_on = [
     module.network
--- a/modules/oke/nodepools.tf
+++ b/modules/oke/nodepools.tf
@@ -10,7 +10,7 @@ resource "oci_containerengine_node_pool" "nodepools" {
   kubernetes_version = var.cluster_kubernetes_version
   name               = var.label_prefix == "none" ? each.key : "${var.label_prefix}-${each.key}"
 
-  freeform_tags = var.freeform_tags["node_pool"]
+  freeform_tags = merge(var.freeform_tags["node_pool"], lookup(each.value, "freeform_tags", {}))
   defined_tags  = merge(var.defined_tags["node_pool"], lookup(each.value, "defined_tags", {}))
 
   node_config_details {
@@ -53,8 +53,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     # allow zero-sized node pools
     size = max(0, lookup(each.value, "node_pool_size", 0))
 
-    freeform_tags = var.freeform_tags["persistent_volume"]
-    defined_tags  = var.defined_tags["persistent_volume"]
+    freeform_tags = var.freeform_tags["node"]
+    defined_tags  = var.defined_tags["node"]
   }
 
   # setting shape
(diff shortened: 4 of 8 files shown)


hyder:
Hi @valireds, 

Thank you for your contribution. Before we can accept and merge your PR, can I please ask you to sign the OCA: https://oca.opensource.oracle.com/?ojr=home?

In the meantime, I'll do the PR review.

main.tf
```diff
@@ -92,7 +93,9 @@ module "bastion" {
   bastion_notification_protocol = var.bastion_notification_protocol
   bastion_notification_topic    = var.bastion_notification_topic
 
+  # tags
   freeform_tags = var.freeform_tags["bastion"]
+  defined_tags  = var.defined_tags["bastion"]
```
hyder, line 98: The bastion module (https://github.com/oracle-terraform-modules/terraform-oci-bastion) doesn't have `defined_tags` yet. Please either
1. Remove it for now and add it for bastion when we've added `defined_tags` in the bastion module.
2. Or implement `defined_tags` in the bastion module. We'll cut a new release for it so we can then use this.

main.tf
```diff
@@ -142,7 +145,9 @@ module "operator" {
   operator_notification_protocol = var.operator_notification_protocol
   operator_notification_topic    = var.operator_notification_topic
 
+  # tags
   freeform_tags = var.freeform_tags["operator"]
+  defined_tags  = var.defined_tags["operator"]
```
hyder, line 150: The operator module (https://github.com/oracle-terraform-modules/terraform-oci-operator) doesn't have `defined_tags` yet. Please either
1. Remove it for now and add it for operator when we've added `defined_tags` in the operator module.
2. Or implement `defined_tags` in the operator module. We'll cut a new release for it so we can then use this.

modules/oke/nodepools.tf
```diff
@@ -53,6 +54,7 @@ resource "oci_containerengine_node_pool" "nodepools" {
     size = max(0, lookup(each.value, "node_pool_size", 0))
 
     freeform_tags = var.freeform_tags["persistent_volume"]
```
hyder, line 56: I think using the persistence_volume tag here was a mistake. It shouldn't have been used here. Please remove the equivalent for `defined_tags` as well.

modules/oke/nodepools.tf
```diff
@@ -11,6 +11,7 @@ resource "oci_containerengine_node_pool" "nodepools" {
   name               = var.label_prefix == "none" ? each.key : "${var.label_prefix}-${each.key}"
 
   freeform_tags = var.freeform_tags["node_pool"]
+  defined_tags  = merge(var.defined_tags["node_pool"], lookup(each.value, "defined_tags", {}))
```
hyder, line 14: Am I right in understanding that you'll be creating a `defined_tags` for each node_pool but with potentially a default one for all of them? Something like this:

```yaml
np1 = { shape = "VM.Standard.E4.Flex", ocpus = 2, memory = 32, node_pool_size = 2, boot_volume_size = 150, label = { app = "frontend", pool = "np1" }, defined_tags = { app = "frontend", pool = "np1" } }

np2 = { shape = "VM.Standard.E4.Flex", ocpus = 2, memory = 32, node_pool_size = 2, boot_volume_size = 150, label = { app = "frontend", pool = "np2" } }
```

If yes, can you also please do the same for freeform_tags?

hyder (changes requested):
Reviewed.

valireds:
> Thank you for your contribution. Before we can accept and merge your PR, can I please ask you to sign the OCA: https://oca.opensource.oracle.com/?ojr=home?

Hi @hyder!
It is a pleasure to contribute to the project! 
I have already signed the OCA in July for the project terraform-oci-oke and I can find myself in the contrib list: 

https://oca.opensource.oracle.com/?ojr=contrib-list

Please tell me if I have to perform further actions to become a contributor! Thanks!

hyder:
> > Thank you for your contribution. Before we can accept and merge your PR, can I please ask you to sign the OCA: https://oca.opensource.oracle.com/?ojr=home?
> 
> Hi @hyder! It is a pleasure to contribute to the project! I have already signed the OCA in July for the project terraform-oci-oke and I can find myself in the contrib list:
> 
> https://oca.opensource.oracle.com/?ojr=contrib-list
> 
> Please tell me if I have to perform further actions to become a contributor! Thanks!

Hi @valireds,
I've found you now (my bad) and we are able to accept and merge your code once the requested changes/clarification in the review are completed.

main.tf
valireds: I removed them for the moment

main.tf
valireds: As for the bastion tags, I removed them for the moment

modules/oke/nodepools.tf
valireds: It could be useful to specify tags for single nodes (well, I am actually using this feature), so I just added the key "node" to the map for both freeform and defined tags

modules/oke/nodepools.tf
valireds: Yes, it is exactly as you said: specifying tags for a particular node pool can override the default ones.
In the latest commit I added the same behavior also for freeform_tags.

modules/oke/nodepools.tf
```diff
@@ -53,8 +53,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     # allow zero-sized node pools
     size = max(0, lookup(each.value, "node_pool_size", 0))
 
-    freeform_tags = var.freeform_tags["persistent_volume"]
-    defined_tags  = var.defined_tags["persistent_volume"]
+    freeform_tags = var.freeform_tags["node"]
```
hyder, line 56: Where do you get the value of node?

terraform.tfvars.example
```diff
@@ -329,6 +333,10 @@ defined_tags = {
       environment = "dev"
       role        = "node-pool"
     }
+    node = {
+      environment = "dev"
+      role        = "instance"
```
hyder, line 338: Instead of instance, a default of "worker" is more appropriate I think.

terraform.tfvars.example
```diff
@@ -179,7 +179,7 @@ image_signing_keys = []
 check_node_active = "all"
 enable_pv_encryption_in_transit = false
 node_pools = {
-  np1 = { shape = "VM.Standard.E4.Flex", ocpus = 1, memory = 16, node_pool_size = 1, boot_volume_size = 150, label = { app = "frontend", pool = "np1" } }
+  np1 = { shape = "VM.Standard.E4.Flex", ocpus = 1, memory = 16, node_pool_size = 1, boot_volume_size = 150, label = { app = "frontend", pool = "np1" }, freeform_tags = { app = "frontend", pool = "np1" }, defined_tags = { app = "frontend", pool = "np1" } }
```
hyder, line 182: Instead of using the same k-v pairs as the node labels, perhaps it would be more useful to have different values for freeform and defined tags.

modules/oke/nodepools.tf
hyder: I see we already have 1 for `node_pool`. What will be the difference between the 2?

docs/terraformoptions.adoc
```diff
@@ -574,6 +574,33 @@ EOT
     }
   }
 }`
+
+|defined_tags
+|Tagging using administrator defined tags. This is specified in the form of a map of maps.
+| e.g.
+`defined_tags = {
+  vcn = {
+    environment = "dev"
```
hyder, line 583: Defined tags need to include the tag namespace on the left in between quotes e.g.

```
    "cn.vcn" = "dev"
```

when "cn" is the tag namespace and "vcn" is the tag key whereas "dev" is the value. Note also that a defined tag can also have a list of values.

Tag namespaces and tag keys must already exist in order for the resources to be successfully created. Additionally, if the tag key is of type list, then the specified value must be a valid value.

As such, I think the default values should be empty as tags would typically be created by administrators.

Can you please:
1. update the example values so that the tag keys are between quotes ""
2. add documentation how to use the tags
3. add error handling in case the tag namespace, keys or values use are invalid

modules/oke/nodepools.tf
valireds: You can apply node_pool tags to their nodes but an user could want to add specific tags to node instances, for example to perform queries on instances using the search engine and for reporting.
Also, I think that this also reflects the behavior of OCI web interface, through which you can specify both node pool tags and node tags.

hyder: I'm not sure I follow. Can you please elaborate a bit more on your use case or what you're trying to achieve with the node tag? 

I agree with the common nodepool tag but I don't quite follow the node one.

valireds: I am simply trying to reproduce the behavior of Oracle Cloud Console, that allows the user to specify tags for nodes of a particular node pool but not for the pool itself. 

For my specific use case, I can just apply the node pool tags to the single nodes, because the most important requirement is to assign a specific tag to resources of type _instance_.
Let me know if I should delete the key `node` and use the `node_pool` instead.

docs/terraformoptions.adoc
valireds: Regarding the error handling, should I use preconditions? If I remember correctly, using them means that the Terraform minimum version will be 1.2 and not 1.0.0

hyder: I've done another round of test and I now understand more where you're going with the node tags and you're in the right direction.

So, similar to _nodepool_ tags, we need to have common and tags for the nodes that are specific to the nodepool. So, https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L56 and https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L57 should read similar as https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L13 and https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L14 respectively except that you'll be extracting the `node` key instead of `node_pool`.

Re: using Terraform 1.2, I'm ok with bumping the minimum Terraform requirement if this means we have a more solid version. We can cut a new minor release to indicate our support for it. Are there any impact on state? I haven't seen any.

modules/oke/nodepools.tf
hyder: I've done another round of test and I now understand more where you're going with the node tags and you're in the right direction.

So, similar to nodepool tags, we need to have common and tags for the nodes that are specific to the nodepool. So, https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L56 and https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L57 should read similar as https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L13 and https://github.com/valireds/terraform-oci-oke/blob/issue-560/modules/oke/nodepools.tf#L14 respectively except that you'll be extracting the node key instead of node_pool.

hyder: I should add: well done for spotting that the nodes needed tags on their own. I initially thought setting tags for the nodepools would be sufficient.

hyder:
Closes #560

hyder:
Hey @valireds,

Any update?

valireds:
Hi @hyder,

Sorry for the delay, but I was trying to implement what we have discussed about. Regarding Terraform version, I am using v1.2.2 to test the template and I don't see any particular issue. 
Regarding the error handling, I tried to write a couple of solutions but it is quite a troublesome activity, mostly because in order to retrieve all the existing defined namespaces + their tags I have to get the infos through 2 data sources, and one of them must cycle on namespace ids.
Even though I managed to prepare a first precondition on the creation of _oci_containerengine_cluster_ resource by analyzing the namespaces of the tenancy and the ones given in input, I think that there is still a lot of work to do to complete a strong solution.
So, could it be possible in this PR to just merge the possibility of using definedTags with documentation and provide a more efficient error handling in another release?

Please consider that in case of a not defined tag, the oci provider is already returning the following error:

```
│ Error: 400-RelatedResourceNotAuthorizedOrNotFound, The following tag namespaces / keys are not authorized or not found: TagNamespace testoke does not exists.
│
│ Suggestion: Please retry or contact support for help with service: Containerengine Cluster
│ Documentation: https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/containerengine_cluster
│ API Reference: https://docs.oracle.com/iaas/api/#/en/containerengine/20180222/Cluster/UpdateCluster
```

which I think is extremely clear and easy to understand and provides all the necessary infos to fix the problem.

hyder:
hi @valireds,

Thanks for trying, much appreciated. I understand it's a bit more involved and I'm ok with moving this forward (pending testing) and adding the error handling in a future release.

Can I request you to please put a section in the documentation though? I think here would be particularly useful: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/docs/configuration.adoc . 

You can use a similar wording like this: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/docs/configuration.adoc#configure-ocir-parameters where we are asking the user to ensure they have already created something. 

I'll test this tomorrow.

modules/oke/nodepools.tf
valireds: This should be fixed in the latest version.

variables.tf
```diff
@@ -960,6 +960,26 @@ variable "freeform_tags" {
...
+      persistent_volume = {}
+      service_lb = {}
+      node_pool = {}
+      node = {}
+    }
+  }
+  description = "Tags to apply to different resources."
+  type = object({
+    vcn      = map(any),
+    oke      = map(map(any))
```
hyder, line 979: `oke = map(any)`  should be sufficient I think.

terraform.tfvars.example
```diff
@@ -291,6 +291,40 @@ freeform_tags = {
       environment = "dev"
       role        = "node-pool"
     }
+    node = {
+      environment = "dev"
+      role        = "instance"
```
hyder, line 296: should this match to defined_tags default and have worker as role instead?

docs/configuration.adoc
```diff
@@ -286,3 +288,8 @@ image_signing_keys = ["ocid1.key.oc1....", "ocid1.key.oc1...."]
 ----
 {uri-terraform-options}#kms-integration[Reference]
 
+== Configure defined tags
+
+The {uri-oci-tags-overview}#[defined tags] *must* first be manually created before using them as input parameters. 
+If a tag is a list type, you must specify one of the predefined options when applying it.
```
hyder, line 294: Can you please give an example for both? Your example from `terraform.tfvars.example` should be good

hyder (changes requested):
Cosmetic changes. I'm going to test now.

terraform.tfvars.example
```diff
@@ -179,7 +179,7 @@ image_signing_keys = []
 check_node_active = "all"
 enable_pv_encryption_in_transit = false
 node_pools = {
-  np1 = { shape = "VM.Standard.E4.Flex", ocpus = 1, memory = 16, node_pool_size = 1, boot_volume_size = 150, label = { app = "frontend", pool = "np1" } }
+  np1 = { shape = "VM.Standard.E4.Flex", ocpus = 1, memory = 16, node_pool_size = 1, boot_volume_size = 150, label = { app = "frontend", pool = "np1" }, freeform_tags = { app = "frontend", pool = "np1" }, defined_tags = { "Operations.CostCenter" = "42" } }
```
hyder, line 182: This defined _tags for specific nodepools worker nodes are not coming through. The one specified for all worker nodes below are coming through.

terraform.tfvars.example
```diff
@@ -291,6 +291,40 @@ freeform_tags = {
...
+    }    
+    service_lb = {
+      "cn.environment" = "dev"
+      "cn.role"        = "load balancer"
+    }
+    node_pool = {
+      "cn.environment" = "dev"
+      "cn.role"        = "node-pool"
+    }
+    node = {
```
hyder, line 324: The defined tags specified here are coming through but the ones specified inside the node pool definition are not coming through

terraform.tfvars.example
hyder: Can you please check again?

hyder (changes requested):
I think we are missing the additional defined tags inside the nodepool

terraform.tfvars.example
valireds: This example sets tags specific to nodepool `np1` but it doesn't apply them to the single nodes because the keys to define them are `freeform_node_tags` and `defined_node_tags`, as you can see from `modules/oke/nodepools.tf:56`

If you want I can add a coalesce function to perform the merge between the specific mapping for the nodes or the one of the node pool, if the first one is not defined. This should ensure the maximum flexibility but let me know how you think it could be solved.

hyder: I see the problem now. It's actually working which is great. However, there is potential for confusion.

For node pools, you are looking up `freeform_tags` and `defined_tags` whereas for worker nodes, you are looking up  `freeform_node_tags` and `defined_node_tags`.

Can you please standardize the tags as follows:

For node pools:
- nodepool_freeform_tags
- nodepool_defined_tags

and worker nodes:
- node_freeform_tags
- node_defined_tags

Also, please update the example in `terraform.tfvars.example`?

One more round and we should be able to merge!

Thanks again for your patience.

valireds: Hi @hyder!
I just updated it, so it should be fine now.

hyder (approved):
@valireds 

Thanks so much for working on this.

PR #563, opened 2022-09-03
fix: Add missing NSG rules for VCN-Native Pod Networking

devoncrouse:
Changed to reflect https://docs.oracle.com/en-us/iaas/Content/ContEng/Concepts/contengnetworkconfig.htm#securitylistconfig:
* Moved `oci_core_network_security_group_security_rule.workers_egress_flannel` to CNI-agnostic `local.workers_egress` for node to node.
* Added `oci_core_network_security_group_security_rule.workers_ingress_npn` for pod to worker.
* Added `oci_core_network_security_group_security_rule.pods_ingress` for CP to pod, worker to pod, and pod to pod (`local.pods_ingress` existed for them but was not yet referenced).
* Added to TCP 6443/12250 to `local.pods_egress` for pod to CP.
* Updated `oci_core_network_security_group_security_rule.pods_egress_internet` from tcp to all protocols.

Signed-off-by: Devon Crouse <devon.crouse@oracle.com>

### DIFF
--- a/modules/network/locals.tf
+++ b/modules/network/locals.tf
@@ -2,6 +2,8 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 locals {
+  # VCN subnet configuration
+  # See https://docs.oracle.com/en-us/iaas/Content/ContEng/Concepts/contengnetworkconfig.htm#vcnconfig
 
   # first vcn cidr
   # pick the first cidr block in the list as this is where we will create the oke subnets
@@ -54,9 +56,11 @@ locals {
     waf_subnet.cidr
   ] : []
 
+  # Security configuration
+  # See https://docs.oracle.com/en-us/iaas/Content/ContEng/Concepts/contengnetworkconfig.htm#securitylistconfig
   # if port = -1, allow all ports
 
-  #control plane seclist
+  # Security List rules for control plane subnet (Flannel & VCN-Native Pod networking)
   cp_egress_seclist = [
     {
       description      = "Allow Bastion service to communicate to the control plane endpoint. Required for when using OCI Bastion service.",
@@ -78,7 +82,8 @@ locals {
       stateless   = false
     }
   ]
-  # control plane
+
+  # Network Security Group egress rules for control plane subnet (Flannel & VCN-Native Pod networking)
   cp_egress = [
     {
       description      = "Allow Kubernetes Control plane to communicate to the control plane subnet. Required for when using OCI Bastion service.",
@@ -114,6 +119,7 @@ locals {
     },
   ]
 
+  # Network Security Group ingress rules for control plane subnet (Flannel & VCN-Native Pod networking)
   cp_ingress = [
     {
       description = "Allow worker nodes to control plane API endpoint communication"
@@ -149,8 +155,16 @@ locals {
     },
   ]
 
-  # workers
+  # Network Security Group egress rules for workers subnet (Flannel & VCN-Native Pod networking)
   workers_egress = [
+    {
+      description      = "Allows communication from (or to) worker nodes.",
+      destination      = local.workers_subnet
+      destination_type = "CIDR_BLOCK",
+      protocol         = local.all_protocols,
+      port             = -1,
+      stateless        = false
+    },
     {
       description      = "Allow ICMP traffic for path discovery",
       destination      = local.anywhere
@@ -185,6 +199,7 @@ locals {
     }
   ]
 
+  # Network Security Group ingress rules for workers subnet (Flannel & VCN-Native Pod networking)
   workers_ingress = [
     {
       description = "Allow ingress for all traffic to allow pods to communicate between each other on different worker nodes on the worker subnet",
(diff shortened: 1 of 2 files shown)


modules/network/locals.tf
```diff
@@ -238,19 +246,35 @@ locals {
...
+      stateless        = false
+    }
   ]
 
   pods_ingress = [
     {
-      description = "Allow worker nodes to access pods.",
+      description = "Kubernetes API endpoint to pod communication.",
       protocol    = local.all_protocols,
       port        = -1,
```
hyder, line 271: Can we be more specific re: protocols and ports?

modules/network/locals.tf
```diff
@@ -238,19 +246,35 @@ locals {
...
       source      = local.cp_subnet,
       source_type = "CIDR_BLOCK",
       stateless   = false
     },
     {
-      description = "Allow Kubernetes Control Plane to communicate with pods.",
+      description = "Allow pods on one worker node to communicate with pods on other worker nodes.",
       protocol    = local.all_protocols,
       port        = -1,
       source      = local.workers_subnet,
```
hyder, line 280: Should the source be the pod subnet? 

Also, can we be more specific re: protocols and ports?

modules/network/nsgs.tf
```diff
@@ -149,19 +149,6 @@ resource "oci_core_network_security_group_security_rule" "workers_egress" {
   count = length(local.workers_egress)
 }
 
-resource "oci_core_network_security_group_security_rule" "workers_egress_flannel" {
```
hyder, line 152: Why is this being removed?

modules/network/nsgs.tf
```diff
@@ -307,6 +294,17 @@ resource "oci_core_network_security_group_security_rule" "workers_healthcheck_in
 
 }
 
+resource "oci_core_network_security_group_security_rule" "workers_ingress_npn" {
```
hyder, line 297: Since it's about pods communication, shouldn't this be covered under the pod NSG instead of the workers? The pods will be using the additional vnics and they should be covered by this rule: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/network/locals.tf#L261

modules/network/nsgs.tf
```diff
@@ -366,14 +364,25 @@ resource "oci_core_network_security_group_security_rule" "pods_egress" {
   count = var.cni_type =="npn" ? length(local.pods_egress) : 0
 }
 
+resource "oci_core_network_security_group_security_rule" "pods_ingress" {
```
hyder, line 367: Ah I think I missed that one. Thanks for picking this up

hyder (commented):
Please clarify the additions/deletions in the PR. We need to ensure they still work for flannel

hyder:
@Djelibeybi Can you please check why @devoncrouse is required to sign OCA?

Djelibeybi:
@devoncrouse you either need to make your OGHO membership public (if you're already a member) or you need to become a member of OGHO and make that membership public once you are.

modules/network/locals.tf
devoncrouse: The rule here didn't change - the description for it vs. the one below seemed flipped based on source subnet. That said, I do think we can make them more specific, e.g. this one is documented but might not be necessary - I'll find out and follow up.

modules/network/nsgs.tf
devoncrouse: It's also needed for NPN - I moved it to the shared locals.workers_egress since the cni_type conditional isn't needed.

modules/network/nsgs.tf
devoncrouse: The description comes from the documentation and could be more clear. e.g. `Pod -> NodePort` or `pod -> hostNetwork pod` would be permitted by this rule. I'll clarify in the description here.

modules/network/locals.tf
devoncrouse: Updating the description to be more clear; this is the other side of the `workers_ingress_npn` change further below, to support cross-host `NodePort -> Pod` or `hostNetwork pod -> pod`.

modules/network/nsgs.tf
hyder: I was going by the docs. Let's try getting the rules more explicitly documented

modules/network/locals.tf
devoncrouse: Updating the description to reflect the use-case for API server calling webhooks served by pods.

modules/network/nsgs.tf
```diff
@@ -366,14 +364,25 @@ resource "oci_core_network_security_group_security_rule" "pods_egress" {
   count = var.cni_type =="npn" ? length(local.pods_egress) : 0
```
hyder, line 364: Can you please add the same condition for the creation of the pod nsg?

devoncrouse: Done.

hyder (approved):
Approved. Thanks a lot for the painstaking verifications. Much appreciated!

PR #568, opened 2022-09-08
fix cp_ingress_additional_cidrs nsg rule

ooraini:
The rule had both icmp and tcp options. Running terraform apply twice always reports changes.


hyder:
Hi @ooraini,

I noticed you closed the PR before we could merge. Is this still an issue?

PR #583, opened 2022-10-02
fix: Ignore lifecycle changes, support Resource Manager variables

### LINKED ISSUE #555: Add the ability to set bastion server and operator server SSH usernames
### Community Note

### Description

Currently the only allowed username for both the bastion and operator servers is `opc`. This assumes that the image used by the servers is an `Oracle Linux` image. However, that's not always the case. Would it be possible to add two new variables `operatoer_username` and `bastion_username` to be able to utilize others images such as `Ubuntu` and be able to set the username to `ubuntu` instead of `opc`?

All initialization scripts that have to be piped through the bastion server that assume the bastion server username is `opc`. These reside in the `modules/external/` directory.

```hcl
```

### LINKED ISSUE #556: `uuid()` in `operator_instance_principal_dynamic_group` causes recreation on every apply
### Community Note

Terraform Version: 1.2.7
Provider Version: 4.87.0
Module Version: 4.2.6

### Affected Resource(s)

oci_identity_policy.operator_instance_principal_dynamic_group

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. 
# Please remove any sensitive information from configuration files before sharing them. 
```

### Expected Behavior

Every time a new plan or apply is made the `oci_identity_policy.operator_instance_principal_dynamic_group` resource should not be recreated since nothing changes about it from a user's perspective.

### Actual Behavior

The `oci_identity_policy.operator_instance_principal_dynamic_group` resource is recreated on every plan/apply cycle because of the usage of `uuid()` in the name.

### Steps to Reproduce

1. Use the module with an operator and bastion servers.
2. Apply this configuration.
3. Plan afterwards and you will see the `oci_identity_policy.operator_instance_principal_dynamic_group` is recreated because of a name change due to the `uuid()` used in the name.

### References

There was a similar issue referring to `defined_tags` that causes recreation, but in this case it's `name` instead. https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/305

devoncrouse:
## Summary
- Ignore lifecycle changes for automatic tagging and dynamic variables/templates
- Support variable names populated by [Resource Manager](https://docs.oracle.com/en-us/iaas/Content/ResourceManager/Concepts/terraformconfigresourcemanager.htm#configvar)
- Use specific compartment for AD listing in case different
- Use stable random suffix for dynamic group name (resolves #556)
- Fall back to provided bastion/operator addresses, SSH user (resolves #555)

## Related
Some lifecycle changes are reported from other modules:
https://github.com/oracle-terraform-modules/terraform-oci-operator/pull/68
https://github.com/oracle-terraform-modules/terraform-oci-operator/pull/69
https://github.com/oracle-terraform-modules/terraform-oci-bastion/pull/50

### Before
Repeated apply:
```
# module.vcn[0].oci_core_service_gateway.service_gateway[0] will be updated in-place
~ resource "oci_core_service_gateway" "service_gateway" {
  ~ defined_tags   = {
      - "Oracle-Tags.CreatedBy" = "user" -> null
      - "Oracle-Tags.CreatedOn" = "2022-10-02T01:54:23.637Z" -> null
    }
```

Repeated destroy:
```
Changes to Outputs:
  - bastion_service_instance_id = "null" -> null
  - fss_id                      = "null" -> null
  - kubeconfig                  = "export KUBECONFIG=generated/kubeconfig" -> null

Destroy complete! Resources: 0 destroyed.
```

### After
Repeated apply:
```
No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

Repeated destroy:
```
Destroy complete! Resources: 59 destroyed.

terraform destroy

Changes to Outputs:

Destroy complete! Resources: 0 destroyed.
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -59,7 +59,7 @@ module "drg" {
 
 module "bastion" {
   source  = "oracle-terraform-modules/bastion/oci"
-  version = "3.1.1"
+  version = "3.1.2"
 
   tenancy_id     = local.tenancy_id
   compartment_id = local.compartment_id
@@ -108,7 +108,7 @@ module "bastion" {
 
 module "operator" {
   source  = "oracle-terraform-modules/operator/oci"
-  version = "3.1.0"
+  version = "3.1.1"
 
 
   # general oci parameters
--- a/main.tf
+++ b/main.tf
@@ -347,11 +347,13 @@ module "extensions" {
   # bastion
   create_bastion_host = var.create_bastion_host
   bastion_public_ip   = local.bastion_public_ip
+  bastion_user        = var.bastion_user
   bastion_state       = var.bastion_state
 
   # operator details
   create_operator                    = var.create_operator
   operator_private_ip                = local.operator_private_ip
+  operator_user                      = var.operator_user
   operator_state                     = var.operator_state
   operator_dynamic_group             = local.operator_instance_principal_group_name
   enable_operator_instance_principal = var.enable_operator_instance_principal
--- a/main.tf
+++ b/main.tf
@@ -400,7 +400,8 @@ module "extensions" {
   nodepool_upgrade_method = var.nodepool_upgrade_method
   node_pools_to_drain     = var.node_pools_to_drain
 
-  debug_mode = var.debug_mode
+  debug_mode        = var.debug_mode
+  update_kubeconfig = var.update_kubeconfig
 
   depends_on = [
     module.bastion,
--- a/main.tf
+++ b/main.tf
@@ -6,7 +6,7 @@ module "vcn" {
   version = "3.5.1"
 
   # general oci parameters
-  compartment_id = var.compartment_id
+  compartment_id = local.compartment_id
   label_prefix   = var.label_prefix
 
   # gateways
@@ -35,12 +35,11 @@ module "vcn" {
 }
 
 module "drg" {
-
   source  = "oracle-terraform-modules/drg/oci"
   version = "1.0.3"
 
   # general oci parameters
-  compartment_id = var.compartment_id
+  compartment_id = local.compartment_id
   label_prefix   = var.label_prefix
 
   # drg parameters
@@ -62,8 +61,8 @@ module "bastion" {
   source  = "oracle-terraform-modules/bastion/oci"
   version = "3.1.1"
 
-  tenancy_id     = var.tenancy_id
-  compartment_id = var.compartment_id
+  tenancy_id     = local.tenancy_id
+  compartment_id = local.compartment_id
 
   label_prefix = var.label_prefix
 
@@ -111,10 +110,10 @@ module "operator" {
   source  = "oracle-terraform-modules/operator/oci"
   version = "3.1.0"
 
-  tenancy_id = var.tenancy_id
 
   # general oci parameters
-  compartment_id = var.compartment_id
+  tenancy_id     = local.tenancy_id
+  compartment_id = local.compartment_id
   label_prefix   = var.label_prefix
 
   # networking
@@ -162,7 +161,7 @@ module "bastionsvc" {
   source = "./modules/bastionsvc"
 
   # general oci parameters
-  compartment_id = var.compartment_id
+  compartment_id = local.compartment_id
   label_prefix   = var.label_prefix
 
   # bastion service parameters
@@ -184,7 +183,7 @@ module "network" {
   source = "./modules/network"
 
   # general oci parameters
-  compartment_id = var.compartment_id
+  compartment_id = local.compartment_id
   label_prefix   = var.label_prefix
 
   # oke networking parameters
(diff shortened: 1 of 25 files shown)


main.tf
```diff
@@ -348,11 +347,13 @@ module "extensions" {
   # bastion
   create_bastion_host = var.create_bastion_host
   bastion_public_ip   = local.bastion_public_ip
+  bastion_user        = var.bastion_user
```
hyder, line 350: How are you ensuring that the bastion can have a different user? The use case is to use the default 'ubuntu' as a user for the ubuntu image. We need to make a change in the bastion module so that an Ubuntu image can also be used.

main.tf
```diff
@@ -348,11 +347,13 @@ module "extensions" {
...
   bastion_state       = var.bastion_state
 
   # operator details
   create_operator                    = var.create_operator
   operator_private_ip                = local.operator_private_ip
+  operator_user                      = var.operator_user
```
hyder, line 356: Ditto for the operator. We need to make a change in the operator module so that an Ubuntu image can also be used.

main.tf
```diff
@@ -400,7 +401,8 @@ module "extensions" {
   nodepool_upgrade_method = var.nodepool_upgrade_method
   node_pools_to_drain     = var.node_pools_to_drain
 
-  debug_mode = var.debug_mode
+  debug_mode        = var.debug_mode
+  update_kubeconfig = var.update_kubeconfig
```
hyder, line 405: I was actually thinking of deleting the local kubeconfig. What's your view on this? keep or do not keep?

modules/extensions/iam.tf
```diff
@@ -17,11 +17,25 @@ locals {
   create_operator_instance_principal_dynamic_group = (var.use_cluster_encryption == true && var.create_policies == true && var.create_bastion_host == true && var.enable_operator_instance_principal == true)
 }
 
+resource "random_id" "dynamic_group_suffix" {
```
hyder, line 20: Awesome. Can you please also implement this in the operator module?

modules/extensions/iam.tf
```diff
@@ -17,11 +17,25 @@ locals {
...
+  keepers = {
+    # Generate a new suffix only when variables are changed
+    label_prefix         = local.dynamic_group_prefix
+    tenancy_id           = var.tenancy_id
+  }
+
+  byte_length = 8
+}
+
 resource "oci_identity_policy" "operator_instance_principal_dynamic_group" {
```
hyder, line 30: I think we need to rename this policy to something more meaningful.

modules/extensions/kubeconfig.tf
```diff
@@ -5,7 +5,21 @@ data "oci_containerengine_cluster_kube_config" "kube_config" {
   cluster_id = var.cluster_id
 }
 
+# Terraform doesn't support conditional dynamic blocks for lifecycle, so resource
+# is repeated with and without it based on update_kubeconfig boolean variable
+# Not ideal, but mitigates reported change on every apply
 resource "local_file" "kube_config_file" {
```
hyder, line 11: As mentioned above, I was actually thinking of getting rid of this. The problem is that this file is overwritten if I use the oke module in a multi cluster scenario. It's also a remnant from when we had a public cluster only. Granted we can still deploy public cluster and users may still want this as a convenience. 

1. What's your view on keeping/not keeping?
2. If we keep, can we please rename by cluster name instead?

main.tf
devoncrouse: It doesn't go that far yet - just connecting config in the module. It could still support a non-default user on a custom OL image this way, or work with a provided bastion host. Future enhancements to bastion will be ready to use with the variables in place.

hyder (changes requested):
feedback inline

main.tf
devoncrouse: I don't use it - in favor of removal.

hyder: Ok, go ahead then.

modules/extensions/kubeconfig.tf
hyder: We agreed to remove this.

main.tf
devoncrouse: We decided to leave for a separate PR.

modules/extensions/kubeconfig.tf
devoncrouse: We decided to leave for another PR.

modules/extensions/iam.tf
devoncrouse: https://github.com/oracle-terraform-modules/terraform-oci-operator/pull/69

main.tf
devoncrouse: Added a note on setting to "opc" for supported images.

main.tf
devoncrouse: Added a note on setting to "opc" for supported images.

hyder (approved):
Approved

PR #587, opened 2022-10-11
Update modules, support base64-encoded SSH key

devoncrouse:
* Update modules: vcn 3.5.2, bastion 3.1.3, operator 3.1.2
* Support base64-encoded SSH, API keys
* Add `provider.tf.example` for convenience


hyder (approved):
approved

PR #588, opened 2022-10-11
fix: Provide empty fallback value for user_id local

devoncrouse:
### Fall back to empty string when user_id and current_user_ocid are both undefined.
```
│ Error: Error in function call
│
│   on .terraform/modules/admin/locals.tf line 10, in locals:
│   10:   user_id = coalesce(var.user_id, var.current_user_ocid)
│     ├────────────────
│     │ var.current_user_ocid is ""
│     │ var.user_id is ""
```
### Update documentation for `provider.tf.example`
image

Links to https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/provider.tf.example


hyder (commented):
Approved

hyder (approved):
Approved

PR #590, opened 2022-10-12
network: use label_prefix in subnets dns_label

### LINKED ISSUE #589: Prefix DNS labels
I'm trying to create my seconds OKE cluster using the module in the same VCN. I got the following error:
```
Error: 400-InvalidParameter, DNS Label workers is a duplicate of one that already exists in the VCN
```

This is because the first cluster already created the subnets with the same DNS labels. The DNS labels are NOT prefixed with the `label_prefix`:
```
resource "oci_core_subnet" "workers" {
  cidr_block                 = local.workers_subnet
  compartment_id             = var.compartment_id
  display_name               = var.label_prefix == "none" ? "workers" : "${var.label_prefix}-workers"
  dns_label                  = "workers"
  prohibit_public_ip_on_vnic = var.worker_type == "private" ? true : false
  route_table_id             = var.worker_type == "private" ? var.nat_route_id : var.ig_route_id
  vcn_id                     = var.vcn_id
}
```

If there are no reasons for it, I will make a PR to change it to `dns_label = var.label_prefix == "none" ? "workers" : "${var.label_prefix}-workers"`

And also for the rest of the subnets.

ooraini:
Signed-off-by: Omar Aloraini <ooraini.dev@gmail.com>

Closes #589

### DIFF
--- a/modules/network/subnets.tf
+++ b/modules/network/subnets.tf
@@ -5,31 +5,43 @@ resource "oci_core_subnet" "cp" {
   cidr_block                 = local.cp_subnet
   compartment_id             = var.compartment_id
   display_name               = var.label_prefix == "none" ? "control-plane" : "${var.label_prefix}-control-plane"
-  dns_label                  = "cp"
+  dns_label                  = var.label_prefix == "none" ? "cp" : "${var.label_prefix}cp"
   prohibit_public_ip_on_vnic = var.control_plane_type == "private" ? true : false
   route_table_id             = var.control_plane_type == "private" ? var.nat_route_id : var.ig_route_id
   security_list_ids          = [oci_core_security_list.control_plane_seclist.id]
   vcn_id                     = var.vcn_id
+
+  lifecycle {
+    ignore_changes = [dns_label]
+  }
 }
 
 resource "oci_core_subnet" "workers" {
   cidr_block                 = local.workers_subnet
   compartment_id             = var.compartment_id
   display_name               = var.label_prefix == "none" ? "workers" : "${var.label_prefix}-workers"
-  dns_label                  = "workers"
+  dns_label                  = var.label_prefix == "none" ? "workers" : "${var.label_prefix}wo"
   prohibit_public_ip_on_vnic = var.worker_type == "private" ? true : false
   route_table_id             = var.worker_type == "private" ? var.nat_route_id : var.ig_route_id
   vcn_id                     = var.vcn_id
+
+  lifecycle {
+    ignore_changes = [dns_label]
+  }
 }
 
 resource "oci_core_subnet" "pods" {
   cidr_block                 = local.pods_subnet
   compartment_id             = var.compartment_id
   display_name               = var.label_prefix == "none" ? "pods" : "${var.label_prefix}-pods"
-  dns_label                  = "pods"
+  dns_label                  = var.label_prefix == "none" ? "pods" : "${var.label_prefix}po"
   prohibit_public_ip_on_vnic = true
   route_table_id             = var.nat_route_id
   vcn_id                     = var.vcn_id
+
+  lifecycle {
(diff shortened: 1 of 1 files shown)


hyder:
Thanks for the PR. Are there any impacts on existing clusters?

ooraini:
@hyder If you did not set `label_prefix` then nothing will change. But, if you are using `label_prefix` then things will not work, because you are not allowed to change the `dns_label` once the subnet is created. 

Not sure how to approach this...

devoncrouse:
Can we lifecycle ignore_changes for `dns_label` on each of the resources, to have this only affect new instances?

ooraini:
The `dns_label` doesn't support hyphens or underscores, only alphanumeric characters are supported. I have used `8` as separator. Another limitation is the maximum length of the `dns_label` is 15. Given the suffix `8workers`, the maximum length for the `label_prefix` is 7, which isn't enough for `production` :/

WDYT?

ooraini:
| full | short |
| -----|-----------|
| workers | wo |
| control plane | cp |
| pods | po |
| public load balancer | pb | 
| internal load balancer | ib |

We could shorted them.

hyder:
I like the shortened labels - they are similar in spirit to the shortened versions of Kubernetes resources. I think that's brilliant. Just concatenate without the 8 though.

Just to confirm, after your changes, if somebody applies this on an existing tfoke created infra, this will not force a deletion and recreation of subnets, cluster, worker nodes. Correct?

ooraini:
@hyder Yes, I just tested the change on an existing cluster and there were no deletions. Thanks to `ignore_changes` tip from @devoncrouse. I'll update suffixes to the shorter version. Without the separator the will look like:

```
productionwo
productioncp
productionpo
productionpb
productionib
```

modules/network/subnets.tf
```diff
@@ -5,56 +5,76 @@ resource "oci_core_subnet" "cp" {
...
+
   count = var.load_balancers == "internal" || var.load_balancers == "both" ? 1 : 0
 }
 
 resource "oci_core_subnet" "pub_lb" {
   cidr_block                 = local.pub_lb_subnet
   compartment_id             = var.compartment_id
   display_name               = var.label_prefix == "none" ? "pub_lb" : "${var.label_prefix}-pub_lb"
-  dns_label                  = "publb"
+  dns_label                  = var.label_prefix == "none" ? "publb" : "${var.label_prefix}pb"
```
hyder, line 69: Are you sure hyphens are not allowed? https://docs.oracle.com/en-us/iaas/Content/Network/Concepts/dns.htm

hyder (changes requested):
I think we should have `labelprefix`-`shorthand`. See if you can manipulate the string to have this format

modules/network/subnets.tf
ooraini: > VCN and subnet labels: Max 15 alphanumeric characters and must start with a letter. Notice that hyphens and underscores are NOT allowed. The value cannot be changed later.

What do you by labelprefix-shorthand?

hyder: So instead of `productionwo`, you'll have `production-wo`

ooraini: It's not allowed, I tested it and I got an error. The documentation confirmed it. We can't use dashes and underscores in the dns label.

hyder:
hey @devoncrouse

Are you ok with this? You asked me to hold on the merge. Do we have your go-ahead on this?

devoncrouse:
lgtm - thanks for addressing the lifecycle changes.

devoncrouse (approved):

hyder (approved):

pojbik:
Is there a reason `dns_label` doesn't accept hyphens ? with this change merged, its restricting our ability to produce unique label_prefix in our automation runs

PR #599, opened 2022-10-25
[WIP] feat: Flexible worker group definition

devoncrouse:
This change adds a new Terraform sub-module (`workergroup`) to organize configuration for OKE worker node provisioning and management.


Djelibeybi (changes requested):
It's also worth using [Markdownlint](https://github.com/markdownlint/markdownlint), [shellcheck](https://www.shellcheck.net/) and [shfmt](https://github.com/mvdan/sh) to check the syntax and structure of your Markdown and shell scripts.

modules/network/outputs.tf
```diff
@@ -38,8 +38,10 @@ output "pod_nsg_id" {
   value = tolist([oci_core_network_security_group.pods[*].id])
 }
 
-output "worker_nsg_id" {
-  value = oci_core_network_security_group.workers.id
+output "pod_nsg_ids" {
```
hyder, line 41: Please add it to the `nsg_ids` instead of a separate output. All the other individual nsg ids should be marked (via comment) as deprecated.

modules/workergroup/README.md
```diff
@@ -0,0 +1,141 @@
...
+worker_group_image_id          = "ocid1.image..." # Required here and/or on group
+worker_group_mode              = "node-pool"
+worker_group_shape             = "VM.Standard.E4.Flex"
+worker_group_ocpus             = 2
+worker_group_memory            = 16
+worker_group_boot_volume_size  = 100
+worker_nsg_ids                 = []
+worker_compartment_id          = "" # Defaults to compartment_id when empty
+worker_group_primary_subnet_id = "" # Defaults to Terraform-managed when empty
+worker_groups                  = [
```
hyder, line 54: It looks like the worker groups are defined in a list. This is different from node pools which is a map. 

Questions:
1. Is there any reason why you selected this?
2. Is it possible to keep map as the same data structure?

The reason we used map for node pools is to facilitate with upgrade of node pools. Although the worker nodes don't have identity, a node pool does have identity. This way, during an upgrade, a user can select exactly which pool to upgrade. With a list, this is not possible.

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+locals {
+  tenancy_id = coalesce(var.tenancy_id, var.tenancy_ocid)
+}
+
+locals {
```
hyder, line 8: Please make a single locals block. If there's a reason they should be in different locals block, then use a different file like we are doing in extensions:

- locals.tf
- templates.tf

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  cluster = lookup({ # Look for existing cluster by OCID, used in subsequent locals blocks
+    for c in lookup(data.oci_containerengine_clusters.wg_clusters, "clusters", []) : c.id => c if c.id == var.cluster_id
+  }, var.cluster_id, {})
+  compartment_id   = coalesce(var.worker_compartment_id, var.compartment_id, var.compartment_ocid, local.tenancy_id)
```
hyder, line 12: What's the difference between `compartment_id` and `compartment_ocid`?

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  region           = coalesce(var.region, "us-ashburn-1")
```
hyder, line 13: The default should be the home_region. Some users may choose not to subscribe to any other regions other than their tenancy's home region.

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  mode             = coalesce(var.mode, "node-pool")
+  size             = max(var.size, 0)
+  shape            = coalesce(var.shape, "VM.Standard.E4.Flex")
+  boot_volume_size = coalesce(var.boot_volume_size, 50)
+  memory           = coalesce(var.memory, 8)
```
hyder, line 18: The default is 16 I think

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  ad_numbers = keys(local.ad_number_to_name)
+  ad_number_to_name = {
+    for ad in data.oci_identity_availability_domains.ad_list.availability_domains :
+    parseint(substr(ad.name, -1, -1), 10) => ad.name
+  }
+
+  # Instance tags - variables + constant
+  defined_tags = merge(coalesce(var.defined_tags, {}), {})
+  freeform_tags = merge(coalesce(var.freeform_tags, {}), {
+    oke_component = "worker",
```
hyder, line 40: Please use `role` instead of `oke_component`.

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  })
+}
+
+locals {
```
hyder, line 44: Again, if there's a reason for them to be in a separate locals block, then please put them in a different file.

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  apiserver_private_endpoint = lookup(element(lookup(local.cluster, "endpoints", []), 0), "private_endpoint")
+  cluster_k8s_version        = lookup(local.cluster, "kubernetes_version", local.k8s_version)
+  k8s_version_length         = length(local.cluster_k8s_version)
+  k8s_version_only           = substr(local.cluster_k8s_version, 1, local.k8s_version_length)
+}
+
+locals {
```
hyder, line 51: as above

modules/workergroup/locals.tf
```diff
@@ -0,0 +1,95 @@
...
+  )
+
+  cluster_networks_output = (
+    length(oci_core_cluster_network.cluster_networks) > 0
+    ? oci_core_cluster_network.cluster_networks
+    : null
+  )
+}
+
+locals {
```
hyder, line 84: as above

modules/workergroup/nodepools.tf
```diff
@@ -0,0 +1,153 @@
+# Copyright 2017, 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+# Dynamic resource block for Node Pool groups defined in worker_groups
+resource "oci_containerengine_node_pool" "nodepools" {
+  for_each = {
```
hyder, line 6: Can you please elaborate on the algorithm behind this?

modules/workergroup/nodepools.tf
```diff
@@ -0,0 +1,153 @@
...
+        availability_domain = ad_iterator.value
+        subnet_id           = var.primary_subnet_id
+      }
+    }
+
+    nsg_ids = lookup(each.value, "worker_nsg_ids", var.worker_nsg_ids)
+
+    # flannel requires cni type only
+    dynamic "node_pool_pod_network_option_details" {
+      for_each = coalesce(var.cni_type, "flannel") == "flannel" ? [1] : []
```
hyder, line 41: In the default node pools, the code to check for cni_type is

```for_each = var.cni_type == "flannel" ? [1] : []```

Can we please make it standard?

File: modules/workergroup/nodepools.tf [hunk 13]
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  # Optimized OKE images
+  dynamic "node_source_details" {
+    for_each = var.image_type == "oke" ? [1] : []
+    content {
+      source_type             = data.oci_containerengine_node_pool_option.np_options.sources[0].source_type
+      boot_volume_size_in_gbs = lookup(each.value, "boot_volume_size", local.boot_volume_size)
+      # check for GPU,A1 and other shapes. In future, if some other shapes or images are added, we need to modify
+      image_id = ((var.image_type == "oke" && length(regexall("GPU|A1", lookup(each.value, "shape", local.shape))) == 0)
```
hyder, line 91: Can we please request this https://github.com/oracle/terraform-provider-oci/issues/1285 is expedited? I don't fancy maintaining this if we have new options or remove existing options in the future

modules/workergroup/variables-iam.tf
```diff
@@ -0,0 +1,45 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+# OCI provider
+# Overrides Resource Manager
```
hyder, line 5: Why is the tenancy_id being passed to the submodule?

modules/workergroup/variables-shape.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "boot_volume_size" {
+  default     = 50
+  description = "Default size in GB for the boot volume of created worker nodes"
+  type        = number
+}
+
+variable "memory" {
+  description = "Default memory in GB for flex shapes"
+  type        = number
+  default     = 8
```
hyder, line 13: Default is 16

modules/workergroup/variables.tf
```diff
@@ -0,0 +1,42 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+variable "create" {
```
hyder, line 4: Please use a meaningful variable name: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/docs/codingconventions.adoc

variables-workergroups.tf
```diff
@@ -0,0 +1,106 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+variable "worker_groups" {
+  default     = []
```
hyder, line 5: We should use map to facilitate upgrades

vars-workergroup.auto.tfvars.example
```diff
@@ -0,0 +1,37 @@
...
+worker_group_image_id         = "ocid1.image..." # Required here and/or on group
+worker_group_mode             = "node-pool"
+worker_group_shape            = "VM.Standard.E4.Flex"
+worker_group_ocpus            = 2
+worker_group_memory           = 16
+worker_group_boot_volume_size = 100
+worker_nsg_ids                 = []
+worker_compartment_id          = "" # Defaults to compartment_id when empty
+worker_group_primary_subnet_id = "" # Defaults to Terraform-managed when empty
+worker_groups = [
```
hyder, line 17: Please use maps to facilitate upgrades

workergroups.tf
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+module "workergroup" {
```
hyder, line 4: It looks like this module is being invoked without any control or dependencies e.g. it should at least wait for the cluster to be ready.

I'm not sure this is the right level to place this module either. Happy to discuss.

hyder (changes requested):
1st round of review completed

modules/workergroup/locals.tf
devoncrouse: `ocid` is what's automatically populated by [Resource Manager](https://docs.oracle.com/en-us/iaas/Content/ResourceManager/Concepts/terraformconfigresourcemanager.htm#configvar) and used by the [OCI Provider](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/terraformproviderconfiguration.htm); this supports either with fallback using a local.

modules/workergroup/nodepools.tf
devoncrouse: Will clarify in comments also: this applies the node_pool resource for elements in worker_groups defined with that mode only, that are also enabled w/ "create" (pending rename based on feedback).

modules/workergroup/nodepools.tf
```diff
@@ -0,0 +1,153 @@
...
+    dynamic "node_pool_pod_network_option_details" {
+      for_each = coalesce(var.cni_type, "flannel") == "flannel" ? [1] : []
+      content {
+        cni_type = "FLANNEL_OVERLAY"
+      }
+    }
+
+    # native requires max pods/node, nsg ids, subnet ids
+    dynamic "node_pool_pod_network_option_details" {
+      for_each = coalesce(var.cni_type, "flannel") == "npn" ? [1] : []
```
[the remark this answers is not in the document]
devoncrouse: Paranoia :) This one actually has validation though, so that's not necessary - will standardize.

modules/workergroup/nodepools.tf
devoncrouse: Agreed... Will see what I can find.

workergroups.tf
devoncrouse: Referencing variables from a module establishes dependencies implicitly, no?

modules/workergroup/clusternetworks.tf
```diff
@@ -0,0 +1,43 @@
...
+
+  placement_configuration {
+    availability_domain = local.ad_number_to_name[lookup(each.value, "placement_ads", local.ad_numbers)[0]]
+    primary_subnet_id   = var.primary_subnet_id
+  }
+
+  lifecycle {
+    ignore_changes = [
+      display_name,
+      time_updated,
```
tariq1890, line 33: Looks like this field raises this warning

```
╷
│ Warning: Redundant ignore_changes element
│ 
│   on modules/workergroup/clusternetworks.tf line 5, in resource "oci_core_cluster_network" "cluster_networks":
│    5: resource "oci_core_cluster_network" "cluster_networks" {
│ 
│ Adding an attribute name to ignore_changes tells Terraform to ignore future changes to the argument in configuration after the object has been created, retaining the value originally configured.
│ 
│ The attribute time_updated is decided by the provider alone and therefore there can be no configured value to compare with. Including this attribute in ignore_changes has no effect. Remove the attribute
│ from ignore_changes to quiet this warning.
```

modules/workergroup/datasources.tf
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022, Oracle Corporation and/or affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+data "oci_identity_availability_domains" "ad_list" {
```
tariq1890, line 4: @devoncrouse This needs to call the the `oci.home` provider

modules/workergroup/clusternetworks.tf
devoncrouse: Fixed; thanks!

devoncrouse:
Converting to draft.

PR #602, opened 2022-10-27
feat: Remove local kubeconfig file creation and `update_kubeconfig` input var

### LINKED ISSUE #601: Avoid writing kubeconfig file
Hi! :wave: 

Thanks for this module! it truly makes it easier to set up an OKE cluster :)

I would like to be able to not write the kubeconfig file, but seems that it's not possible to avoid that. Am I missing something?

To give more context. We execute terraform using Terraform cloud, which is executed in ephemeral workers, so the file disappears on every run, instead, we store the content of the kubeconfig in Hasicorp's Vault.

This ends that on every run, terraform creates a new resource, so we are always in a drift.

If it's not possible, would be something that you would be interested in? I could submit a PR if required. In the end, I think that where you store the kubeconfig data is up to the user.

slok:
Fixes #601 

This PR removes the creation of a local kubeconfig file and adds instructions so, users select how it wants to manage the kubeconfig information data.

Also removes the `update_kubeconfig` flag as isn't required anymore.

Note: I don't know where to add the changelog changes for unreleased ones.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,6 +7,65 @@ All notable changes to this project are documented in this file.
 
 The format is based on {uri-changelog}[Keep a Changelog].
 
+== 4.2.14 (not released)
+=== Changes
+* feat: Removed local kubeconfig file creation and `update_kubeconfig` input variable by @slok in #602
+
+== 4.2.13
+=== Changes
+* network: use label_prefix in subnets dns_label by @ooraini in #590
+* chore: cosmetic changes to load balancer subnet names and labels by @hyder in #592
+* feat: upgraded operator module to 3.1.4 by @hyder in #598
+
+== 4.2.12
+=== Changes
+* fix: Provide empty fallback value for user_id local by @devoncrouse in #588
+
+== 4.2.11
+=== Changes
+* Update modules, support base64-encoded SSH key by @devoncrouse in #587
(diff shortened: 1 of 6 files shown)


devoncrouse:
Thanks for submitting the PR - this looks good to me. Can I please ask you to sign the Oracle Contributor Agreement referenced in the earlier comment? Once verified, we'll be able to accept and merge this.

> Note: I don't know where to add the changelog changes for unreleased ones.

It looks like we haven't kept up with this lately. If you'd like to include an entry, you can add an unreleased section with placeholder version, to be associated with a release when next updated like in [this commit](https://github.com/oracle-terraform-modules/terraform-oci-vcn/commit/52158fbc20599bd8027e7b0c9551ac091b2de2a9#diff-bc34fd17a77bda3e2cbef0e783a3ee58a970a9bb65f9476cfdd27c8e78831a93).

slok:
I signed yesterday (almost took me more time than the PR itself xD), and it's under review right now.

l also just added all the missing changelog entries (with the information from GitHub releases) and added this PR entry as unreleased.

hyder:
@slok: LGTM. We can merge once the OCA is signed and approved.

slok:
@devoncrouse @hyder All set! Thanks :)

hyder (approved):
Approved. Thank you for your contribution

PR #606, opened 2022-11-09
feat: Support disabled DNS record assignment for OKE subnets

devoncrouse:
# Proposed changes

1. Add a new `assign_dns` boolean variable with default of `true` to match current behavior. When `false`, the optional `dns_label` values for OKE subnets will be unset, resulting in disabled DNS resolution of hostnames within the subnet. This also provides one option to avoid issues with `label_prefix` when the DNS names are unused; other changes for added flexibility will be considered separately.

References:
* [core_vcn#dns_label](https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/core_vcn#dns_label)
* [oci_core_subnet#dns_label](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/core_subnet#dns_label)

```
assign_dns = false
```

> DNS Domain Name: DNS isn’t enabled for this VCN
> DNS Domain Name: DNS isn’t enabled for this Subnet

2. Don't create nodepools when variable is unspecified
3. Use defaults in tfvars example

## How has these changes been tested?

Manually, on new and existing tfstate with null and default values.

### DIFF
--- a/terraform.tfvars.example
+++ b/terraform.tfvars.example
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 # Identity and access parameters
-api_fingerprint      = ""
+api_fingerprint = ""
 # api_private_key      = <<EOT
 #-----BEGIN RSA PRIVATE KEY-----
 #content+of+api+key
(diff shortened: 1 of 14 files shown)


terraform.tfvars.example
```diff
@@ -34,6 +34,9 @@ ssh_private_key_path = "~/.ssh/id_rsa"
 # ssh_public_key_path = "~/.ssh/id_rsa.pub"
 
 # networking
+
+assign_dns = true
```
hyder, line 38: please move this together with the vcn input variables

variables.tf
```diff
@@ -938,6 +944,16 @@ variable "enable_gatekeeper" {
 }
 
 variable "gatekeeeper_version" {
```
hyder, line 946: Please remove it. It will only destroy a null_resource. The installation would already have been done if enabled.

hyder (changes requested):
Should we also wait until the dns changes in the vcn module are completed and merged?

hyder (approved):
Approved

PR #607, opened 2022-11-14
WIP: feat: Flexible worker group definition

devoncrouse:
This change adds a new Terraform sub-module (`workergroup`) to organize configuration for OKE worker node provisioning and management.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,6 @@ provider.tf
 generated/**
 
 # visual code
- **/.vscode/*
+**/.vscode/*
 
 .terraform.lock.hcl
--- a/locals.tf
+++ b/locals.tf
@@ -1,29 +1,39 @@
-# Copyright 2017, 2021 Oracle Corporation and/or affiliates.
+# Copyright 2017, 2022 Oracle Corporation and/or affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 locals {
-  tenancy_id     = coalesce(var.tenancy_id, var.tenancy_ocid)
+  tenancy_id = coalesce(var.tenancy_id, var.tenancy_ocid)
   compartment_id = coalesce(
     var.compartment_id, var.compartment_ocid,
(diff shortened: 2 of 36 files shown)


pojbik:
I know this is a WIP but we were experimenting with workergroup config. Should this module output here - https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/feat-wg/outputs-workergroups.tf#L16  be referring to `module.workergroup.worker_groups_enabled` instead of `module.workergroup.enabled_worker_groups`

devoncrouse:
> Should this module output here - https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/feat-wg/outputs-workergroups.tf#L16 be referring to `module.workergroup.worker_groups_enabled` instead of `module.workergroup.enabled_worker_groups`

Yes - this is fixed, thank you.

pojbik:
Are there any plane to define structured types for `worker_groups` var instead of `any`
```
variable "worker_groups" {
  default     = {}
  description = "Tuple of OKE worker groups where each key maps to the OCID of an OCI resource, and value contains its definition"
  type        = any
}
```
Currently has above. But I see that var mapping might be a problem when we try to model these vars
similar scenario is described here - https://github.com/hashicorp/terraform/issues/22185#issuecomment-515639220

devoncrouse:
> Are there any plans to define structured types for `worker_groups` var instead of `any`

Hi @pbikki - thanks for weighing in on the proposed structure. I was indeed looking at structured types for this early on, though we ultimately decided that the `any` type used by existing `node_pools` historically still yields the most flexibility. Using a map instead of list, we avoid changes/handling for reordered elements, and in practice have not experienced issues like the one referenced, since map keys must be string and values are consistently `map(any)` in this case.

This flexibility can be useful in interesting ways, e.g. supporting custom fields on group definitions - with no effect on group creation in the module, but persisted in TF state for e.g. programmatic use elsewhere, or added context from a `terraform show`.

Having said that, if you're finding that this structure isn't accommodating a desired configuration in some way, I'm interested to know more of the specifics and open to exploring options.

PR #615, opened 2022-11-30
Output Endpoints

gotsysdba:
Small change to output endpoints... justification:

Deploying the stack via Oracle Resource Manager/Market Place.  Terraform calls an Ansible playbook as part of the deployment for config mgmt of the cluster and requires access to the control plane (no operator is deployed).  

Access to the CP can be made through Resource Manager Private Endpoints, but to do so, the private endpoint IP must be known, for example:

```
data "oci_resourcemanager_private_endpoint_reachable_ip" "orm_pe_reachable_ip" {
  count               = var.orm_install ? 1 : 0
  private_endpoint_id = oci_resourcemanager_private_endpoint.orm_pe[0].id
  private_ip          = split(":", module.oke.cluster_endpoints[0].private_endpoint)[0]
}
```

This will provide an IP (`data.oci_resourcemanager_private_endpoint_reachable_ip.orm_pe_reachable_ip[0].ip_address`) to access the private endpoint of the CP via the ORM/MP private endpoint.

### DIFF
--- a/modules/oke/outputs.tf
+++ b/modules/oke/outputs.tf
@@ -12,3 +12,7 @@ output "cluster_kms_dynamic_group_id" {
 output "nodepool_ids" {
   value = zipmap(values(oci_containerengine_node_pool.nodepools)[*].name, values(oci_containerengine_node_pool.nodepools)[*].id)
 }
+
+output "endpoints" {
+  value = oci_containerengine_cluster.k8s_cluster.endpoints
+}
--- a/outputs.tf
+++ b/outputs.tf
@@ -84,3 +84,8 @@ output "bastion_service_instance_id" {
 output "fss_id" {
   value = var.create_fss == true ? module.storage[0].fss_id : null
 }
+
+output "cluster_endpoints" {
+  description = "Endpoints for the Kubernetes cluster"
+  value       = module.oke.endpoints
+}


hyder:
Hi @gotsysdba, 

Thanks again for your contribution. I think I need a bit more context here. Can you please explain your use case? I'm not sure I fully grasp it.

gotsysdba:
@hyder
We're building an Oracle Resource Manager (ORM) stack which once the infrastructure is provisioned will kick off an Ansible playbook (abbr. examples):

```
module "oke" {
...
}

resource "null_resource" "ansible_k8s_apply" {
  provisioner "local-exec" {
    command = "ansible-playbook ${abspath(local.ansible_home)}/k8s_apply.yaml"
  }
  depends_on = [ module.oke.nodepool_ids ]
}
```

This playbook will connect to the OKE control plane and do some config mgmt (deploy the ingress service, create secrets, etc).  All stuff that could be done by the operator, but we want to keep the infra to a minimum.

As we allow the user to choose if the OKE CP endpoint is public/private and what CIDRs can connect, we can't guarantee that our playbook will have access unless we connect to the OKE PRIVATE_ENDPOINT (`oci ce cluster create-kubeconfig ... --kube-endpoint PRIVATE_ENDPOINT`) from the CP subnet.

We accomplish this by creating a ORM Private Endpoint in the CP Subnet tied to the CP's Private IP address:

```
resource "oci_resourcemanager_private_endpoint" "orm_pe" {
  compartment_id = local.compartment_ocid
  display_name   = format("%s-orm-pe", local.label_prefix)
  description    = "Private Endpoint for Resource Manager to Control Plane"
  vcn_id         = module.oke.vcn_id
  subnet_id      = module.oke.subnet_ids["cp"]
}

data "oci_resourcemanager_private_endpoint_reachable_ip" "orm_pe_reachable_ip" {
  private_endpoint_id = oci_resourcemanager_private_endpoint.orm_pe[0].id
  private_ip          = split(":", module.oke.cluster_endpoints[0].private_endpoint)[0]
}
```

Once that is established, the `kubeconfig` file is updated with the ORM PE IP address (via the playbook) and connectivity for config mgmt is guaranteed from ORM.

In the end, we just need to get at what that private_endpoint IP address is (module.oke.cluster_endpoints), without protocol or port, to link the ORM PE with it. My proposed PR is to output all the endpoints and let the main module manipulate the output per needs (to cater for any unforeseen use case).

devoncrouse (approved):

PR #624, opened 2022-12-14
Node pool correctly ignore sub-sub-object

12345ieee:
Correct terraform syntax to ignore just the needed parts of `node_metadata` and `node_config_details`

Before the change the whole sub-objects were ignored, leading to inability to change a nodepool size via terraform.

Fixes: 269d3fdd89

### DIFF
--- a/modules/oke/nodepools.tf
+++ b/modules/oke/nodepools.tf
@@ -121,8 +121,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     ignore_changes = [
       kubernetes_version,
       defined_tags, # automatic tagging after apply
-      node_metadata["user_data"], # templated cloud-init
-      node_config_details["placement_configs"], # dynamic placement configs
+      node_metadata, # templated cloud-init
+      node_config_details[0].placement_configs, # dynamic placement configs
       node_source_details # dynamic image lookup
     ]
   }


hyder:
@devoncrouse Can you also please check the placement_configs?

devoncrouse:
:eyes:

modules/oke/nodepools.tf
```diff
@@ -121,8 +121,8 @@ resource "oci_containerengine_node_pool" "nodepools" {
     ignore_changes = [
       kubernetes_version,
       defined_tags, # automatic tagging after apply
-      node_metadata["user_data"], # templated cloud-init
-      node_config_details["placement_configs"], # dynamic placement configs
+      node_metadata[0].user_data, # templated cloud-init
```
devoncrouse, line 124:
```suggestion
      node_metadata.user_data, # templated cloud-init
```
>│ Error: Unsupported attribute
│ 
│   on modules/oke/nodepools.tf line 124, in resource "oci_containerengine_node_pool" "nodepools":
│  124:       node_metadata[0].user_data, # templated cloud-init
│ 
│ Can't access attributes on a primitive-typed value (string).

Small adjustment for the map field on this one

12345ieee: It worked the way I did when I tested it, but it probably depends on how many `node_metadata`s there are.
May as well ignore the whole `node_metadata`.

devoncrouse:
This looks good to me, including the `node_metadata` adjustment mentioned above. Good catch.

modules/oke/nodepools.tf
devoncrouse: I'm good with ignoring `node_metadata` here.

hyder: sounds good. I'll merge when completed.

hyder (dismissed):
Please add the whole node_metadata to ignore

12345ieee:
Here we go.

devoncrouse (approved):

PR #626, opened 2022-12-27
fix: Correct Internet Gateway creation rule

und3f:
The Internet Gateway is not created if the Load Balancer subnet is PRIVATE and Control Plane is PUBLIC. This PR is fixing this issue

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -10,7 +10,7 @@ module "vcn" {
   label_prefix   = var.label_prefix
 
   # gateways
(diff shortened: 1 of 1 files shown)


hyder (approved):
Well spotted @und3f! After you've signed the OCA and it's been accepted, we can merge. 

In the meantime, thanks for your contribution and happy holidays.

und3f:
Joined main Oracle GitHub org, changed my membership to public. Waiting for OCA app to recheck my membership

PR #627, opened 2023-01-03
fix: add preferred_loadbalancer validation

und3f:
In case if user provision OKE with internal LB and doesn't set perferred_load_balancer variable the terraform will fail during provision with unintuitive message:

> Error: 400-InvalidParameter, Invalid subnetId: subnetId must not be blank
> │ Suggestion: Please update the parameter(s) in the Terraform config as per error message Invalid subnetId: subnetId must not be blank
> │ Documentation: https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/containerengine_cluster 
> │ API Reference: https://docs.oracle.com/iaas/api/#/en/containerengine/20180222/Cluster/CreateCluster 
> │ Request Target: POST https://containerengine.us-ashburn-1.oci.oraclecloud.com/20180222/clusters 
> │ Provider version: 4.102.0, released on 2022-12-14. This provider is 2 Update(s) behind to current. 
> │ Service: Containerengine Cluster 
> │ Operation Name: CreateCluster 
> │ OPC request ID: ....
> │ 
> │ 
> │   with module.oke.module.oke.oci_containerengine_cluster.k8s_cluster,
> │   on .terraform/modules/oke/modules/oke/cluster.tf line 11, in resource "oci_containerengine_cluster" "k8s_cluster":
> │   11: resource "oci_containerengine_cluster" "k8s_cluster" {
> │ 

With the fix terraform would fail on a plan phase:
> ╷
> │ Error: Resource precondition failed
> │ 
> │   on ../../modules/oke/cluster.tf line 81, in resource "oci_containerengine_cluster" "k8s_cluster":
> │   81:       condition     = var.cluster_subnets[local.lb_subnet] != ""
> │     ├────────────────
> │     │ local.lb_subnet is "pub_lb"
> │     │ var.cluster_subnets is map of string with 5 elements
> │ 
> │ preferred_load_balancer references unexisting load balancer.

Also found that modules/oke/locals.tf contains unused variable `lb_subnet`, so I simplified `service_lb_subnet_ids` expression by using `local.lb_subnet`

### DIFF
--- a/modules/oke/cluster.tf
+++ b/modules/oke/cluster.tf
@@ -71,11 +71,16 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
       defined_tags  = lookup(var.defined_tags,"service_lb",{})
     }
 
-    service_lb_subnet_ids = var.preferred_load_balancer == "public" ? [var.cluster_subnets["pub_lb"]] : [var.cluster_subnets["int_lb"]]
+    service_lb_subnet_ids = [var.cluster_subnets[local.lb_subnet]]
   }
 
   lifecycle {
     ignore_changes = [defined_tags, freeform_tags, cluster_pod_network_options]
+
+    precondition {
+      condition     = var.cluster_subnets[local.lb_subnet] != ""
+      error_message = "Preferred load balancer references unexisting load balancer. Please check variables preferred_load_balancer and load_balancers."
+    }
   }
 
   vcn_id = var.vcn_id


modules/oke/cluster.tf
```diff
@@ -71,11 +71,16 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
...
-    service_lb_subnet_ids = var.preferred_load_balancer == "public" ? [var.cluster_subnets["pub_lb"]] : [var.cluster_subnets["int_lb"]]
+    service_lb_subnet_ids = [var.cluster_subnets[local.lb_subnet]]
   }
 
   lifecycle {
     ignore_changes = [defined_tags, freeform_tags, cluster_pod_network_options]
+
+    precondition {
+      condition     = var.cluster_subnets[local.lb_subnet] != ""
+      error_message = "preferred_load_balancer references unexisting load balancer."
```
hyder, line 82: Please provide a more intuitive message instead.

hyder (changes requested):
Minor changes and then we can merge. This looks good. Is it possible to add multiple preconditions?

und3f:
Thanks, working on the error message.
Yes, it is possible to specify multiple preconditions

modules/oke/cluster.tf
und3f: What do u thing about
> Preferred load balancer references unexisting load balancer. Please check variables preferred_load_balancer and load_balancers.

hyder: You should guide them to what is wrong in their configuration e.g. ```preferred_load_balancer``` cannot be an empty string. Otherwise, it's quite meaningless as there's no load balancer created with this module.

und3f: `preferred_load_balancer` is not an empty string.

User may archive such configuration if he specify `load_balancers = "internal"` and doesn't specify preferred_load_balancer -- such configuration creates only internal load balancer, but `preffered_load_balancer` is pointing to the default value `public`. The problem is (and I've experienced it when tried module at first) that you may not be aware o the `preferred_load_balancer` variable at all.

So the oke module fails to find the public lb subnet.
I've tried to point a user to the variables `preferred_load_balancer` and `load_balancers` as it will help him to find what is triggering an error.

hyder (approved):

gotsysdba:
For reference:
Users of Oracle Resource Manager will need to stay at v4.4.1 if incorporating this module into a stack.  ORMs latest supported Terraform is **v1.1.0** and [preconditions and postconditions](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#preconditions-and-postconditions) for resources, data sources, and outputs was introduced in Terraform **v1.2.0**.

The precondition in [modules/oke/cluster.tf](https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/627/files#diff-c3643c3d4123fe59ea39d5ad0a5d2e3a8f91c50329c0a0e77c5d1d593d8f6d11) will break ORM stacks which have not explicitly set the module to version 4.4.1 (or earlier) with similar errors:

```text
Error: Invalid type for provider on modules/oke/main.tf
The given provider configuration is for a different provider type.

Error: Unsupported block type on modules/oke/modules/oke/cluster.tf line 80, in resource "oci_containerengine_cluster" "k8s_cluster":
80:     precondition {
Blocks of type "precondition" are not expected here.
```

devoncrouse:
> ORMs latest supported Terraform is **v1.1.0**

It looks like **v1.2.x** may be available very soon. If that changes, we can consider removing the precondition until it is.

hyder:
> > ORMs latest supported Terraform is **v1.1.0**
> 
> It looks like **v1.2.x** may be available very soon. If that changes, we can consider removing the precondition until it is.

Thanks for letting us know @gotsysdba and apologies for the trouble. 

@devoncrouse Going forward, we should also put up a compatibility matrix: Terraform, ORM

gotsysdba:
> It looks like **v1.2.x** may be available very soon. If that changes, we can consider removing the precondition until it is.

Terraform v1.2.x is now available in ORM; 4.4.2 and 4.5.2 tested successfully against it.

Issue #628, opened 2023-01-07
OKE nodepool is limited to 32 characters


christophercorn:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

An OKE nodepool is limited to 32 chars. 

var.label_prefix can sometimes consume most of those chars, leaving little room for the remainder of the node pool name as defined in each key name of var.node_pools. 

for a more descriptive nodepool name we should have the ability to drop the var.label_prefix name specifically for the oci_containerengine_node_pool resource. this way, we can maintain the functionality of  var.label_prefix  that is used across this terraform module.

### New or Affected Resource(s)

oci_containerengine_node_pool

### Potential Terraform Configuration

```hcl
# Copy-paste any Terraform configurations for how the requested feature may be used. 
```

devoncrouse:
Resolved by #630.

PR #630, opened 2023-01-07
feat: Optionally ignore the label prefix in node pool names

christophercorn:
https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/628

An OKE nodepool is limited to 32 chars.

var.label_prefix can sometimes consume most of those chars, leaving little room for the remainder of the node pool name as defined in each key name of var.node_pools.

for a more descriptive nodepool name we should have the ability to drop the var.label_prefix name specifically for the oci_containerengine_node_pool resource. this way, we can maintain the functionality of var.label_prefix that is used across this terraform module.

to validate this change:
```
module "oke" {
  source = "oracle-terraform-modules/oke/oci"
  ignore_label_prefix_in_node_pool_names = true  
  ....
}
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -264,20 +264,21 @@ module "oke" {
   admission_controller_options                            = var.admission_controller_options
 
   # oke node pool parameters
-  kubeproxy_mode                  = var.kubeproxy_mode
-  max_pods_per_node               = var.max_pods_per_node
-  node_pools                      = var.node_pools
-  node_pool_name_prefix           = var.node_pool_name_prefix
-  node_pool_image_id              = var.node_pool_image_id
-  node_pool_image_type            = var.node_pool_image_type
-  node_pool_os                    = var.node_pool_os
-  node_pool_os_version            = var.node_pool_os_version
-  node_pool_timezone              = var.node_pool_timezone
-  enable_pv_encryption_in_transit = var.enable_pv_encryption_in_transit
-  use_node_pool_volume_encryption = var.use_node_pool_volume_encryption
-  node_pool_volume_kms_key_id     = var.node_pool_volume_kms_key_id
-  cloudinit_nodepool              = var.cloudinit_nodepool
-  cloudinit_nodepool_common       = var.cloudinit_nodepool_common
+  kubeproxy_mode                         = var.kubeproxy_mode
+  max_pods_per_node                      = var.max_pods_per_node
+  node_pools                             = var.node_pools
+  node_pool_name_prefix                  = var.node_pool_name_prefix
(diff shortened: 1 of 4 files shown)


hyder:
Why not use the value ```none``` in your case for the label prefix? It will achieve the same effect.

christophercorn:
> Why not use the value `none` in your case for the label prefix? It will achieve the same effect.

 Label_prefix is used widely by many other resources in the terraform-oci-oke module. Changing label_prefix will affect many other resources that don’t have a small 32 char limitation. The suggested implementation only affects node pool resource name.

hyder:
> > Why not use the value `none` in your case for the label prefix? It will achieve the same effect.
> 
> Label_prefix is used widely by many other resources in the terraform-oci-oke module. Changing label_prefix will affect many other resources that don’t have a small 32 char limitation. The suggested implementation only affects node pool resource name.

So you want to keep the label_prefix for the other resources but keep it optional for node pools in case the name is too long?

christophercorn:
> > > Why not use the value `none` in your case for the label prefix? It will achieve the same effect.
> > 
> > 
> > Label_prefix is used widely by many other resources in the terraform-oci-oke module. Changing label_prefix will affect many other resources that don’t have a small 32 char limitation. The suggested implementation only affects node pool resource name.
> 
> So you want to keep the label_prefix for the other resources but keep it optional for node pools in case the name is too long?

Yes, that would be very helpful.

christophercorn:
Can someone review this and possible approve?

hyder:
Hi, sorry for the delay. 

@devoncrouse and I have been discussing this and we will drop the `label_prefix` in the next major release.

christophercorn:
@hyder  do you know when 5.0 will be released?

hyder:
> @hyder do you know when 5.0 will be released?

We just cut the 1st beta release for 5.x. @devoncrouse does it include dropping the `label_prefix` ?

devoncrouse (approved):
We'll be deprecating `label_prefix` in 5.x, but we can add this to 4.x in the meantime.

christophercorn:
> We'll be deprecating `label_prefix` in 5.x, but we can add this to 4.x in the meantime.

thanks @devoncrouse

hyder:
No, not yet unfortunately. We just released the 2nd beta yesterday.

lodotek:
> We'll be deprecating `label_prefix` in 5.x, but we can add this to 4.x in the meantime.

Can i find some notes regarding this deprecation? This is causing me frustration when trying to use the latest RC.

PR #633, opened 2023-01-12
fix: Correct formatting and some unreferenced vars

devoncrouse:
### Commit summary
- Run `terraform fmt` for all `.tf` files in project
- Run `tflint` at project root (non-recursive; other warnings in sub-modules should be resolved with planned cleanup separately)


devoncrouse:
Discussed with @hyder and applying this separately.

PR #635, opened 2023-01-26
Secondary VNIC for IP/CN, compatible defaults for block volume/AD

devoncrouse:
- Add experimental configuration for secondary VNICs
- Use correct defaults by mode for block volume config
- Use Kubernetes v1.25 by default
- Use consistent worker image lookup and variables
- Add precondition for better error message with missing image_id
- Fix worker pool resource display names with removed label_prefix

### DIFF
--- a/docs/terraformoptions.adoc
+++ b/docs/terraformoptions.adoc
@@ -799,25 +799,25 @@ node_pools = {
 }
 |{}
 
-|node_pool_image_id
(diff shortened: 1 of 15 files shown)


hyder (commented):
When you add a secondary vnic, you need to run the `oci-network-config`. See https://docs.oracle.com/en-us/iaas/oracle-linux/oci-utils/index.htm#oci-network-config.

I don't see this command being run. Am I missing something?

devoncrouse:
> you need to run the `oci-network-config`

The host network can be set up other ways as well - this just exposes configuration for the created resources without being prescriptive.

Issue #172, opened 2020-06-14
Add schema for resource manager


hyder:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

https://docs.cloud.oracle.com/en-us/iaas/Content/ResourceManager/Concepts/terraformconfigresourcemanager.htm

devoncrouse:
https://docs.oracle.com/en-us/iaas/Content/ResourceManager/Concepts/terraformconfigresourcemanager_topic-schema.htm

devoncrouse:
Added examples to 5.x [user guide](https://oracle-terraform-modules.github.io/terraform-oci-oke) in https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/703

PR #640, opened 2023-02-02
Enhanced network/policy config and sub-module refactor

devoncrouse:
This includes a number of enhancements for the 5.x branch:
- Organize module and variable declarations into files by role
- Remove node pool implementation in `oke` submodule superseded by worker_pools
- Remove deprecated `label_prefix`, use generated `state_id` for unique naming and tags
- Support defined tags for resources and optional [tag-based access control](https://docs.oracle.com/en-us/iaas/Content/Tagging/Tasks/managingaccesswithtags.htm)
- Support disabled cluster creation to include network/policy resources only
- Support cluster/worker/bastion/operator creation with existing network resources
- Use NSG references for rule definitions between components
- Support extended configuration for additional load balancer NSG rules
- Consolidate IAM policy creation into submodule with additional configuration
- Internalize and simplify bastion and operator components
- Resolve tflint warnings for e.g. unused variables
- Consolidate data sources and logic for AD, image, and other resource reflection
- Support policies and deployment for cluster autoscaler with node pools
- Support flexible cloud-init user configuration for workers/bastion/operator
- Clean definition and await completion of cloud-init for bastion/operator
- Rename `cp` -> `control plane`, `(control_plane|worker|bastion)_type` vars -> `*_is_public` booleans
- Separate `extensions` with Helm from `utilities` submodule

Validation is pending for some extended features along with updated documentation and examples, to be reflected under issues for the 5.x milestone.


module-cluster.tf
```diff
@@ -0,0 +1,68 @@
...
+  service_lb_subnet_id    = lookup(module.network.subnet_ids, var.preferred_load_balancer == "public" ? "pub_lb" : "int_lb")
+  services_cidr           = var.services_cidr
+  vcn_id                  = local.vcn_id
+
+  # OKE cluster
+  cluster_kms_key_id = var.cluster_kms_key_id
+  cluster_name       = var.cluster_name
+  defined_tags       = lookup(var.defined_tags, "cluster", {})
+  freeform_tags      = lookup(var.freeform_tags, "cluster", {})
+  image_signing_keys = var.image_signing_keys
```
hyder, line 39: Where possible, try grouping related input variables together in a subsection. Also, let's specify the control that enable/disable the behaviour first before specifying the parameters. Move them down if you need to.

e.g. `image_signing_keys` will be related to `use_signed_images`. So, let's specify a subsection for them:

```
## kms
use_signed_images = var.use_signed_images
image_signing_keys = var.image_signing_keys
```

Then, if the module, we create the variables in the same order. It then becomes easy to check if we are missing any

module-iam.tf
```diff
@@ -0,0 +1,79 @@
...
+    var.create_kms_policy == "always",
+    var.create_kms_policy == "auto" && anytrue([
+      coalesce(var.worker_volume_kms_key_id, "none") != "none",
+      coalesce(var.cluster_kms_key_id, "none") != "none",
+    ])
+  ])
+}
+
+# Default IAM sub-module implementation for OKE cluster
+module "policy" {
```
hyder, line 36: please rename the module to iam

module-network.tf
```diff
@@ -0,0 +1,187 @@
...
+data "oci_core_route_tables" "all" {
+  compartment_id = local.compartment_id
+  state          = "AVAILABLE"
+  vcn_id         = local.vcn_id
+}
+
+locals {
+  vcn_id = (var.create_vcn ? one(module.vcn[*].vcn_id)
+    : coalesce(var.vcn_id, "none") != "none" ? var.vcn_id
+    : one(one(data.oci_core_vcns.all[*].virtual_networks)[*].id)
```
hyder, line 39: this is assuming that a vcn with the same display name will be the one you're looking for. However, as mentioned above, `display_name` is not unique.

hyder, line 39: this is assuming that a vcn with the same display name will be the one you're looking for. However, as mentioned above, `display_name` is not unique.

module-network.tf
```diff
@@ -0,0 +1,187 @@
...
+}
+
+data "oci_core_vcn" "oke" {
+  count  = var.create_vcn ? 0 : 1
+  vcn_id = var.vcn_id
+}
+
+data "oci_core_nat_gateways" "all" {
+  compartment_id = local.compartment_id
+  display_name   = var.nat_gateway_display_name
```
hyder, line 18: as for vcn, `display_name` is not unique

module-network.tf
```diff
@@ -0,0 +1,187 @@
...
+  state          = "AVAILABLE"
+  vcn_id         = local.vcn_id
+}
+
+data "oci_core_internet_gateways" "all" {
+  compartment_id = local.compartment_id
+  display_name   = var.internet_gateway_display_name
```
hyder, line 25: as for vcn, display_name is not unique

module-network.tf
devoncrouse: That's true - it's making a best effort at that point to find something available rather than fail (e.g. in a lot of cases there may just be 1) - should we require instead?

module-network.tf
devoncrouse: Responded in other comments - we can remove this if we decide to make strict i.e. must provide vcn_id explicitly if create is disabled.

module-network.tf
hyder: Yes, we should fail it. A vcn_id is something specific, it means users know what they are doing whereas looking by name, there's a potential to create something in vcn where it was not intended to be. More importantly, when you look up by name, sort order is not guaranteed and even if you sort it alphanumerically, it may not end up being the vcn the user intended the cluster to go to.

hyder: Yup, as mentioned above. Either the user accepts creating the vcn here or provide a valid vcn ocid. By valid, this means they have the necessary permission etc.

module-workers.tf
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) 2022, 2023 Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+locals {
+  worker_count_expected = coalesce(one(module.workers[*].worker_count_expected), 0)
+
+  # Distinct list of compartments for enabled worker pools
+  worker_compartments = distinct(compact([
```
hyder, line 8: Using different compartments make things more difficult in my view: permissions, debugging in the console. I know there are users who roll this way, especially because they want to restrict access to compute or they want to cost by compartment. 

In those cases, if users want isolation for this kind of purposes, it is better they stand up a different cluster.

module-network.tf
devoncrouse: Done.

module-network.tf
devoncrouse: Removed as with VCN fallback.

module-network.tf
devoncrouse: Done.

module-workers.tf
devoncrouse: I agree with you, but I know it's being done in some cases.

module-workers.tf
```diff
@@ -0,0 +1,111 @@
...
+
+  # Cluster
+  apiserver_private_host = try(split(":", one(module.cluster[*].endpoints.private_endpoint))[0], "")
+  cluster_ca_cert        = local.cluster_ca_cert
+  cluster_dns            = var.cluster_dns
+  cluster_id             = coalesce(var.cluster_id, one(module.cluster[*].cluster_id))
+  kubernetes_version     = var.kubernetes_version
+
+  # Worker pools
+  worker_pool_enabled = var.worker_pool_enabled
```
hyder, line 45: use the imperative mode e.g. create_workers

module-cluster.tf
```diff
@@ -0,0 +1,72 @@
...
+
+module "cluster" {
+  count          = var.create_cluster ? 1 : 0
+  source         = "./modules/cluster"
+  compartment_id = local.compartment_id
+  state_id       = random_id.state_id.id
+
+  # Network
+  cni_type                = var.cni_type
+  control_plane_nsg_ids   = setunion(var.control_plane_nsg_ids, var.create_nsgs ? [module.network.cp_nsg_id] : [])
```
hyder, line 26: Should we make the name of this variable match that in the OCI Console/documentation e.g Kubernetes API endpoint ? It's bit wordy but we can find a short form.

modules/bastion/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+# Common
+variable "compartment_id" { type = string }
+variable "state_id" { type = string }
+
+# Bastion
+variable "assign_dns" { type = bool }
+variable "availability_domain" { type = string }
+variable "image_id" { type = string }
+variable "nsg_ids" { type = list(string) }
+variable "public" { type = bool }
```
hyder, line 13: please keep the same variable names as in bastion's variables. This way it'll be easier to update

data-common.tf
```diff
@@ -0,0 +1,19 @@
...
+resource "random_id" "state_id" {
+  byte_length = 6
+}
+
+data "oci_identity_availability_domains" "all" {
+  compartment_id = local.compartment_id
+}
+
+locals {
+  # Stable availability domain selection
```
hyder, line 13: Is this selecting the AD or just generating a map for selection? Please clarify the purpose of this. It looks useful and reusable so perhaps other modules can benefit from this.

data-images.tf
```diff
@@ -0,0 +1,42 @@
...
+
+  image_ids = merge({
+    x86_64   = [for k, v in local.parsed_images : k if v.arch == "x86_64"]
+    aarch64  = [for k, v in local.parsed_images : k if v.arch == "aarch64"]
+    oke      = [for k, v in local.parsed_images : k if v.image_type == "oke" && v.k8s_version == local.k8s_version_only]
+    platform = [for k, v in local.parsed_images : k if v.image_type == "platform"]
+    gpu      = [for k, v in local.parsed_images : k if v.is_gpu]
+    nongpu   = [for k, v in local.parsed_images : k if !v.is_gpu]
+    }, {
+    for k, v in local.parsed_images : "${v.os} ${split(".", v.os_version)[0]}" => k...
```
hyder, line 40: Blimey, this is cool but apparently buried deep in Terraform docs. Adding this here for future reference. https://gist.github.com/reegnz/7de97f4dd70da6320e6bf86eb883cfdd

modules/bastion/compute.tf
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2019, 2023 Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+locals {
```
hyder, line 4: Can you please keep the same file organization that we are using in https://github.com/oracle-terraform-modules/terraform-oci-bastion?

module-bastion.tf
```diff
@@ -0,0 +1,61 @@
...
+    ? var.bastion_image_id : element(tolist(setintersection([
+      lookup(local.image_ids, format("%s %s", var.bastion_image_os, split(".", var.bastion_image_os_version)[0]), null),
+      local.image_ids.nongpu, local.image_ids.x86_64,
+    ]...)), 0)
+  )
+}
+
+module "bastion" {
+  count          = var.create_bastion ? 1 : 0
+  source         = "./modules/bastion"
```
hyder, line 20: Remote submodules such as vcn, bastion and operator are reusable modules and not just for this project.

If there are enhancements you wish to bring to them, then they should equally apply for the remote and should be implemented there first instead of embedding this locally. We moved away from embedding locally so we don't have to maintain the code in 2 different places.

CONTRIBUTORS.adoc
```diff
@@ -4,6 +4,7 @@ OWNERS # have admin access and can merge code to main:
 
 - @hyder
 - @markxnelson
+- @devoncrouse
```
hyder, line 7: We can remove this file. We have the contribution history in git.

module-bastion.tf
```diff
@@ -0,0 +1,61 @@
...
+  state_id       = random_id.state_id.id
+  compartment_id = local.compartment_id
+
+  # Bastion
+  assign_dns          = var.assign_dns
+  availability_domain = coalesce(var.bastion_availability_domain, lookup(local.ad_numbers_to_names, local.ad_numbers[0]))
+  image_id            = local.bastion_image_id
+  nsg_ids             = concat(var.bastion_nsg_ids, var.create_nsgs ? [module.network.bastion_nsg_id] : [])
+  public              = var.bastion_type == "public"
```
hyder, line 29: Internal/inner module variable names should have the same. Please check the coding conventions for naming variables:

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/docs/codingconventions.adoc#31-case-type-files-names

Perhaps we should rename this to something like `assign_public_ip` ?

variables-common.tf
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017, 2023 Oracle Corporation and/or its affiliates.
+# Copyright (c) 2022, 2023 Oracle Corporation and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 locals {
```
hyder, line 4: Variables files should contain only variable definitions. Please move locals into a locals.tf

hyder: Also consider moving all locals here into 1 file and document the subsections. Otherwise, it's tedious to hunt for a local variable

terraform-oci-oke-kix.code-workspace
```diff
@@ -0,0 +1,20 @@
+{
```
hyder, line 1: This file should not have been committed. Please remove

module-cluster.tf
```diff
@@ -0,0 +1,72 @@
...
+  control_plane_subnet_id = lookup(module.network.subnet_ids, "cp")
+  control_plane_type      = var.control_plane_type
+  pods_cidr               = var.pods_cidr
```
hyder, line 29: If the CNI is NPN, what should this cidr range be?

module-extensions.tf
```diff
@@ -0,0 +1,69 @@
...
+  gatekeeper_version = var.gatekeeper_version
+
+  # Service account
+  create_service_account               = var.create_service_account
+  service_account_name                 = var.service_account_name
+  service_account_namespace            = var.service_account_namespace
+  service_account_cluster_role_binding = var.service_account_cluster_role_binding
+
+  # Worker node readiness
+  check_node_active   = var.check_node_active
```
hyder, line 55: I like "readiness". It's inline with Kubernetes terminology. Let's rename this variable to `check_node_readiness`

module-iam.tf
```diff
@@ -0,0 +1,85 @@
...
+
+  autoscaler_compartments = local.autoscaler_compartments
+  worker_compartments     = local.worker_compartments
+
+  providers = {
+    oci.home = oci.home
+  }
+}
+
+moved {
```
hyder, line 67: Since we are making breaking changes, we can remove the `moved` block right?

module-iam.tf
```diff
@@ -0,0 +1,85 @@
...
+  from = module.policy
+  to   = module.iam
+}
+
+output "availability_domains" {
```
hyder, line 72: Similar to locals, I think it would be good to group the outputs in 1 file. In this example, the `local.ad_numbers_to_names` is defined in data-common.tf but the output is here. 

Let's put all the outputs in 1 file and document the sections. Outputs should be either limited to just enough that makes the module reusable and allow other users to build on this module directly or to find more details using the API if they need to.

module-network.tf
```diff
@@ -0,0 +1,142 @@
...
+module "vcn" {
+  count          = var.create_vcn ? 1 : 0
+  source         = "oracle-terraform-modules/vcn/oci"
+  version        = "3.5.3"
+  compartment_id = coalesce(var.network_compartment_id, local.compartment_id)
+  defined_tags   = var.defined_tags["vcn"]
+  freeform_tags  = var.freeform_tags["vcn"]
+
+  attached_drg_id              = var.drg_id != null ? var.drg_id : (var.create_drg ? module.drg[0].drg_id : null)
+  create_internet_gateway      = !(var.load_balancers == "internal" && !var.create_bastion && var.control_plane_type == "private")
```
hyder, line 32: set the `var.create_bastion == true` to improve readability

module-network.tf
```diff
@@ -0,0 +1,142 @@
...
+  lockdown_default_seclist     = var.lockdown_default_seclist
+  nat_gateway_public_ip_id     = var.nat_gateway_public_ip_id
+  nat_gateway_route_rules      = var.nat_gateway_route_rules
+  vcn_cidrs                    = local.vcn_cidrs
+  vcn_dns_label                = var.assign_dns ? var.vcn_dns_label : null
+  vcn_name                     = var.vcn_name
+}
+
+module "drg" {
+  count          = var.create_drg || var.drg_id != null ? 1 : 0
```
hyder, line 46: set `var.create_drg == true`

module-network.tf
```diff
@@ -0,0 +1,142 @@
...
+module "network" {
+  source           = "./modules/network"
+  state_id         = random_id.state_id.id
+  compartment_id   = coalesce(var.network_compartment_id, local.compartment_id)
+  defined_tags     = lookup(var.defined_tags, "network", {})
+  freeform_tags    = lookup(var.freeform_tags, "network", {})
+  tag_namespace    = var.tag_namespace
+  use_defined_tags = var.use_defined_tags
+
+  allow_node_port_access       = var.allow_node_port_access
```
hyder, line 71: For this set of access rules, they are either ingress or egress. Should we indicate these in the names?

module-network.tf
```diff
@@ -0,0 +1,142 @@
...
+  value       = one(module.drg[*].drg_id)
+}
+
+output "vcn_id" {
+  description = "VCN ID"
+  value       = local.vcn_id
+}
+
+output "network_security_rules" {
+  value = var.output_network_detail ? module.network.network_security_rules : null
```
hyder, line 141: Let's have a `detailed_output` variable that covers everything. No need for a specific one for network.

If it's false, then we give a minimal output.

module-operator.tf
```diff
@@ -0,0 +1,74 @@
...
+
+  operator_image_id = (var.operator_image_type == "custom"
+    ? var.operator_image_id : element(tolist(setintersection([
+      lookup(local.image_ids, format("%s %s", var.operator_image_os, split(".", var.operator_image_os_version)[0]), null),
+      local.image_ids.nongpu, local.image_ids.x86_64,
+    ]...)), 0)
+  )
+}
+
+module "operator" {
```
hyder, line 23: As above for bastion. It's a reusable module.

module-workers.tf
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) 2022, 2023 Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+locals {
+  worker_count_expected = coalesce(one(module.workers[*].worker_count_expected), 0)
+
+  # Distinct list of compartments for enabled worker pools
+  worker_compartments = distinct(compact([
+    for k, v in var.worker_pools : lookup(v, "compartment_id", local.compartment_id)
```
hyder, line 9: If I'm reading this right, you are first looking for the `compartment_id` in each worker pool. If it's not present, try a general `worker_compartment_id`, which if not present try `compartment_id`, which if not present, try `compartment_ocid`. The last case I assume is catch for resource manager.

module-workers.tf
```diff
@@ -0,0 +1,111 @@
...
+
+  # Worker pools
+  worker_pool_enabled = var.worker_pool_enabled
+  worker_pool_mode    = var.worker_pool_mode
+  worker_pool_size    = var.worker_pool_size
+  worker_pools        = var.worker_pools
+
+  # Workers
+  assign_dns            = var.assign_dns
+  assign_public_ip      = var.worker_type == "public"
```
hyder, line 52: Internal and external variables. Let's use the same name.

variables-workers.tf
```diff
@@ -0,0 +1,232 @@
...
+    condition     = contains(["iptables", "ipvs"], var.kubeproxy_mode)
+    error_message = "Accepted values are iptables or ipvs."
+  }
+}
+
+#
+# Workers: instance
+#
+
+variable "worker_block_volume_type" {
```
hyder, line 105: Can you please also add the ability to configure the vpu?

variables-workers.tf
```diff
@@ -0,0 +1,232 @@
...
+    shape            = "VM.Standard.E4.Flex",
+    ocpus            = 2,
+    memory           = 16,
+    boot_volume_size = 50
+  }
+  description = "Default shape of the created worker instance."
+  type        = map(any)
+}
+
+variable "worker_cloudinit" {
```
hyder, line 160: Is this the default cloud init?

variables-common.tf
```diff
@@ -16,6 +16,12 @@ locals {
   )
 }
 
+variable "timezone" {
```
hyder, line 19: Please prefix with `default_worker`

module-workers.tf
```diff
@@ -0,0 +1,111 @@
...
+  block_volume_type     = var.worker_block_volume_type
+  cloudinit             = var.worker_cloudinit
```
hyder, line 54: How are you allowing for multiple cloud-inits?

provider.tf.example
```diff
@@ -1,18 +1,20 @@
 provider "oci" {
-  fingerprint         = var.api_fingerprint
-  private_key         = var.api_private_key
-  region              = var.region
-  tenancy_ocid        = local.tenancy_id
-  user_ocid           = local.user_id
-  config_file_profile = var.config_file_profile
+  config_file_profile  = var.config_file_profile
```
hyder, line 2: `config_file_profile` should be optional

variables-bastion.tf
```diff
@@ -0,0 +1,94 @@
...
+}
+
+variable "bastion_image_os" {
+  default     = "Oracle Linux"
+  description = "Bastion image operating system name when bastion_image_type = 'platform'."
+  type        = string
+}
+
+variable "bastion_image_os_version" {
+  default     = "8"
```
hyder, line 63: Default should be Autonomous Linux 7.9.

variables-bastion.tf
```diff
@@ -0,0 +1,94 @@
...
+  description = "Whether to use a platform or custom image for the created bastion instance. When custom is set, the bastion_image_id must be specified."
+  type        = string
+  validation {
+    condition     = contains(["custom", "platform"], var.bastion_image_type)
+    error_message = "Accepted values are custom or platform"
+  }
+}
+
+variable "bastion_image_os" {
+  default     = "Oracle Linux"
```
hyder, line 57: Add validation. Value can be Oracle Linux, Ubuntu, SUSE, AlmaLinux, Rocky Linux

variables-cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+  type        = bool
+}
+
+variable "cluster_name" {
+  default     = "oke"
+  description = "The name of oke cluster."
+  type        = string
+}
+
+variable "control_plane_type" {
```
hyder, line 16: Should we rename this variable so it has the same terminology as in the console/docs?

variables-cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+
+variable "use_signed_images" {
+  description = "Whether to enforce the use of signed images. If set to true, at least 1 RSA key must be provided through image_signing_keys."
+  default     = false
+  type        = bool
+}
+
+variable "image_signing_keys" {
+  description = "A list of KMS key ids used by the worker nodes to verify signed images. The keys must use RSA algorithm."
+  type        = set(string)
```
hyder, line 57: Is it possible to demand a set of keys if `use_signed_images` is set to `true`?

variables-cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+  default     = []
+}
+
+variable "control_plane_nsg_ids" {
```
hyder, line 61: Group this variable with other `control_plane*` variables

variables-cluster.tf
```diff
@@ -0,0 +1,93 @@
...
+  default     = "public"
+  description = "The preferred load balancer subnets that OKE will automatically choose when creating a load balancer. valid values are public or internal. if 'public' is chosen, the value for load_balancers must be either 'public' or 'both'. If 'private' is chosen, the value for load_balancers must be either 'internal' or 'both'."
+  type        = string
+  validation {
+    condition     = contains(["public", "internal"], var.preferred_load_balancer)
+    error_message = "Accepted values are public or internal."
+  }
+}
+
+variable "output_cluster_detail" {
```
hyder, line 89: As for network.

variables-extensions.tf
```diff
@@ -0,0 +1,179 @@
+# Copyright (c) 2017, 2023 Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+variable "check_node_active" {
```
hyder, line 4: As mentioned above, let's rename this to `check_node_readiness`

variables-extensions.tf
```diff
@@ -0,0 +1,179 @@
...
+}
+
+# Oracle Container Image Registry (OCIR)
+variable "email_address" {
+  default     = "none"
+  description = "The email address used for OCIR."
+  type        = string
+}
+
+variable "secret_id" {
```
hyder, line 34: I think we need to rename this too: maybe `authtoken_id`?

modules/cluster/cluster.tf
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2017, 2023 Oracle Corporation and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
+
+locals {
+  roles = ["cluster", "persistent_volume", "service_lb"]
+
+  defined_tags = { for role in local.roles : role => merge(
```
hyder, line 7: Please document this

modules/workers/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+variable "assign_dns" { type = bool }
+variable "assign_public_ip" { type = bool }
+variable "cni_type" { type = string }
+variable "pod_nsg_ids" { type = list(string) }
+variable "pod_subnet_id" { type = string }
+variable "worker_nsg_ids" { type = list(string) }
+variable "worker_subnet_id" { type = string }
+
+# Worker pools
+variable "worker_pool_enabled" { type = bool }
```
hyder, line 32: imperative mode

modules/workers/instanceconfig.tf
```diff
@@ -6,21 +6,17 @@ resource "oci_core_instance_configuration" "workers" {
...
   display_name   = each.key
+  defined_tags   = each.value.defined_tags
+  freeform_tags  = each.value.freeform_tags
 
   instance_details {
     instance_type = "compute"
 
-
-
     launch_details {
```
hyder, line 15: I don't see agent configuration in the instance config

module-cluster.tf
devoncrouse: The value still exists but is ignored with VCN-Native - consistent with the API. Added clarification.

data-images.tf
devoncrouse: Added comment to clarify w/ reference to https://developer.hashicorp.com/terraform/language/expressions/for#grouping-results

module-cluster.tf
devoncrouse: Just for reference, this would be e.g. `endpoint_config_nsg_ids` or just `endpoint_nsg_ids` - it's the only NSG reference used in this sub-module. This could be valid, but "control plane" seems like a better fit with other variables and resource names (e.g. the subnet) and rules pertain to the OKE control plane as well as the k8s endpoint, so leaning towards this.

variables-workers.tf
devoncrouse: Good idea, but let's do this in another change; there are other storage-related configs that would be worth surfacing as well, but warrant explicit testing and possible preconditions.

variables-common.tf
devoncrouse: This is used for workers, bastion, and operator instances - updated the description.

provider.tf.example
devoncrouse: It is optional - the default profile will be used when the var is unset, same as if not included in the provider block.

variables-bastion.tf
devoncrouse: Validation exists in the form of a precondition on matched `image_id` - the bastion in particular is pretty flexible as it's just used for SSH. If no image is matched, the precondition will yield an error that calls out this value.

modules/workers/instanceconfig.tf
devoncrouse: This hasn't been implemented/tested yet. It'd be a good addition, but let's do it in a separate change.

variables-cluster.tf
devoncrouse: Variable validations may only refer to themselves; I added a precondition in `cluster.tf`.

module-network.tf
devoncrouse: Variable naming and context of logical operators should provide readability with booleans - comparison with static values is unnecessary and may even cause problems with [unexpected implementation semantics](https://medium.com/@xtg/terraform-boolean-evaluation-b866b528e90b). If there's any type ambiguity (e.g. when using map lookups) then wrapping in `tobool` is a good idea/harmless, but a bool var used directly is otherwise fine on its own.

module-operator.tf
devoncrouse: The operator in this context is a single Compute instance that is specialized for OKE with installed packages + kubeconfig, and highly coupled to other components with connected NSGs, subnet creation, input/output variables, consistent tag definition, and IAM policies. There's not enough here IMO to warrant the overhead of maintaining it outside of the project and keeping in sync. `5.x` is making it easier to apply a partial set of components, e.g. using existing VCN/subnets with cluster/NSG creation disabled - it'd be possible to apply the operator component on its own with an existing `cluster_id` this way, though I'd question why anyone should need to. I don't see any other references to the project.

variables-common.tf
devoncrouse: It can be opinionated, and I think there are good reasons to colocate sections when they are closely related. There are always exceptions, but this also isn't visible outside of the module and doesn't impact usage or presentation for users one way or another. I do agree that variables should be isolated though, and only use locals in this context when strictly derived from one/more in the same file to be used in place of them, as is the case with this fallback logic here - the vars shouldn't be used directly outside of the file.
In many (I'd argue most) cases, a pattern of organizing strictly on block type ends up even more disorganized with [tight coupling/low cohesion](https://en.wikipedia.org/wiki/Coupling_(computer_programming)) and some pretty long files to scroll through, e.g.  in `4.x` [modules/extensions/templates.tf](https://raw.githubusercontent.com/oracle-terraform-modules/terraform-oci-oke/c67a0a17a25f42c3793e3c5ee2e0aef8fe3c70fd/modules/extensions/templates.tf), [modules/extensions/locals.tf](https://raw.githubusercontent.com/oracle-terraform-modules/terraform-oci-oke/c67a0a17a25f42c3793e3c5ee2e0aef8fe3c70fd/modules/extensions/locals.tf), [modules/network/locals.tf](https://raw.githubusercontent.com/oracle-terraform-modules/terraform-oci-oke/c67a0a17a25f42c3793e3c5ee2e0aef8fe3c70fd/modules/network/locals.tf), [modules/network/nsgs.tf](https://raw.githubusercontent.com/oracle-terraform-modules/terraform-oci-oke/c67a0a17a25f42c3793e3c5ee2e0aef8fe3c70fd/modules/network/nsgs.tf), etc.
If a file can cleanly encapsulate local values sourced from/used by resources nearby and their outputs, it cuts down on needing to hop around looking for them when making a change and on the number of files affected. No solution is perfect, and there's always Find in Files/grep :shrug:

module-iam.tf
devoncrouse: Point taken on these definitions belonging to `iam` and corrected. We can discuss further in the other related thread.

module-workers.tf
devoncrouse: That's correct, and will be documented with worker pools. I think we can drop the last fallback if #172 is completed.

module-workers.tf
devoncrouse: Do you consider it preferable to keep the `worker_type` var this way? IMO it's already binary, has an ambiguous name, and depends on validation - I'd lean towards matching these on the `assign_public_ip` boolean.

module-workers.tf
devoncrouse: Workers/operator allow specification via `list(map(string))` of MIME multi-parts - the [variable descriptions](https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/640/files#diff-cc4febfbdbdf88d60319cc80a7dba73411d1effd41defdcc8d922a5410915ff3R150) and [precondition validations](https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/640/files#diff-ebb04e6baff3b5ecfb921f4f51dbaeadc6bcebb02c26415c27fef07ffba550d3R100-R142) reference cloud-init docs for the schema and configuration, and additional examples will be included with our own documentation.

variables-cluster.tf
devoncrouse: As with `control_plane_nsg_ids`, this would end up being e.g. `endpoint_config_is_public_ip_enabled`. Aside from being kind of long, it would be a departure from the term used elsewhere that is slightly broader in the cases where we're also referring to the OKE control plane and not just k8s endpoint. Open to considering this more for `5.x`, but I'd propose doing so in a separate change from the file organization here and including all references.

variables-bastion.tf
devoncrouse: I think we should prefer consistency with supported worker images using an OL8 platform image by default and minimal implementation. There will be a good case for some to disable this in favor of the external module for advanced functionality, and this should be easily accomplished by providing its output IP to the optional var here. As with the operator (even more-so), the basic need for most is met by a single lightweight Compute instance, and this relieves the module from passing through a dozen variables not directly related to its purpose + needing to test those input variations. I would've preferred the service if not for that session time limitation...

module-network.tf
devoncrouse: Possibly, though others e.g. `allow_worker_ssh_access` for worker ingress also include creation of bastion egress. Let's consider in another change separate from file organization.

variables-workers.tf
```diff
@@ -0,0 +1,232 @@
...
+  default     = "paravirtualized"
+  description = "The default block volume attachment type for Instance Configurations."
+  type        = string
+  validation {
+    condition     = contains(["iscsi", "paravirtualized"], var.worker_block_volume_type)
+    error_message = "Accepted values are 'iscsi' or 'paravirtualized'."
+  }
+}
+
+variable "worker_node_labels" {
```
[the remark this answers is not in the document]
devoncrouse: Looked into this - I wouldn't disagree if we were talking about a few variables, but the reality is that anything can be defined at the pool level, so they're _all_ "defaults". The variable descriptions call this out; I'm afraid prefixing the names will make them extremely long without adding much value, since the whole paradigm is to allow override, and behavior is described with each (e.g. merge vs. when unspecified).

variables-workers.tf
devoncrouse: Yes, clarified description for behavior.

variables-extensions.tf
devoncrouse: More planned for extensions; let's do this separate from this change.

module-bastion.tf
devoncrouse: 1. Same question as with `worker_type` below - would this be better as an e.g. `assign_public_ip` boolean?
2. I don't personally feel it's necessary to require a full match between inner/outer of a sub-module so long as the difference is only a redundant prefix for inner, e.g. `nsg_ids` and `image_id` for `bastion_nsg_ids` and `operator_image_id` in their respective sub-modules aren't creating any confusion and reduce eye strain. If those sub-modules were to be used directly then having the prefix on every single var would be kind of weird, though no one else sees their spec here; it's just aesthetics for maintainers.

modules/bastion/variables.tf
devoncrouse: Renamed to `is_public` for consistency.

module-workers.tf
devoncrouse: Updated to `worker_is_public` bool as we discussed.

module-bastion.tf
devoncrouse: Updated `worker_type` -> `worker_is_public` as discussed.

variables-common.tf
devoncrouse: Discussed and in agreement.

module-operator.tf
devoncrouse: Discussed and in agreement for now - may revisit in a subsequent PR w/ enhancements to related projects.

variables-bastion.tf
devoncrouse: Discussed and in agreement for now - will revisit use of bastion module in a subsequent PR.

module-bastion.tf
devoncrouse: Discussed and in agreement. Will updated coding conventions.

module-bastion.tf
devoncrouse: Discussed and in agreement for now - will revisit in a subsequent PR.

hyder (approved):
Approved

PR #653, opened 2023-03-09
fix: No public kube_config datasource for private endpoint

devoncrouse:
- Disable public kube_config source with private endpoint


hyder (approved):
Approved

PR #655, opened 2023-03-09
fix: Add compute-management-family to cluster autoscaler policy

devoncrouse:
- Add `compute-management-family` policy used by `oci` cloud provider


hyder (approved):
approved

PR #656, opened 2023-03-09
fix: Worker drain evaluation

devoncrouse:
- Still pending implementation; this change fixes a variable reference issue causing errors in some cases.


hyder:
If this is still being implemented, please put it in draft mode. 

If it's ready, please elaborate the use case i.e. under which circumstances the drain is causing errors

hyder (approved):
Approved

PR #659, opened 2023-03-09
fix: Update AD selection for instance config mode

devoncrouse:
- Use fixed AD selection for `cluster-network` mode when multi-AD


hyder (approved):
Approved

PR #660, opened 2023-03-09
fix: 5.x Update NSG handling, optionally force creation

devoncrouse:
- Support `create_nsgs_always` to force creation e.g. when applying only network resources and bastion.
- Minor fix for worker `nsg_ids` default/pool configuration.
- fix: No utilities for disabled cluster, operator
- fix: Update DRG attachment specification
- fix: Update NSG handling, optionally force creation
- fix: Update operator, extension lifecycle change detection
- style: Update .gitignore
- fix: Create NSG rule for operator to CP if create_nsgs_always
- fix: Dynamic NSG rules
- fix: Output ssh_to_bastion with provided address
- fix: NPN NSG rule conditional, add extra preconditions


module-network.tf
```diff
@@ -83,6 +83,7 @@ module "network" {
   create_bastion               = var.create_bastion
   create_fss                   = var.create_fss
   create_nsgs                  = var.create_nsgs
```
hyder, line 85: what's the difference between `create_nsgs` and `create_nsgs_always`?

devoncrouse: This is similar to the subnets `never`/`auto`/`always` use-case, and I added a comment that we might consider aligning on that for NSGs.

NSGs are created by default when called for, e.g. operator NSG when operator is enabled, Pods with NPN cluster, etc. In cases where network resources are created separately and re-used, `create_nsgs_always` forces creation even when the related components are disabled.

modules/network/nsg-pods.tf
hyder, line 1:
``` "Allow ALL egress from pods to internet" : {
      protocol = local.all_protocols, port = local.all_ports, destination = local.anywhere, destination_type = local.rule_type_cidr,
      enabled  = var.allow_pod_internet_access,
    },```

This should be explicitly controlled IMO.

modules/network/nsg-workers.tf
hyder, line 1: SSH to workers should go through bastion even though the workers are public

hyder: [Line 53-](https://github.com/devoncrouse/terraform-oci-oke/blob/fix-nsgs/modules/network/nsg-workers.tf#L53)[Line 64](https://github.com/devoncrouse/terraform-oci-oke/blob/fix-nsgs/modules/network/nsg-workers.tf#L64)

hyder (changes requested):
changes related to ssh

hyder (approved):
approved

PR #662, opened 2023-03-09
fix: Re-trigger await cloud-init on bastion/operator replacement

devoncrouse:
- Re-trigger null resource to await completion of cloud-init when operator/bastion hosts are replaced.


hyder (approved):
approved

PR #663, opened 2023-03-09
fix: Only add cluster DNS when provided, or determined automatically

devoncrouse:
- Leave user data variable unset when not provided an explicit value - determined automatically.


hyder (approved):
Approved.

PR #665, opened 2023-03-14
fix: Honor changes to pool size for CN

devoncrouse:
Ignored lifecycle for block keys had the block ignored, including `size`.

### DIFF
--- a/modules/workers/clusternetworks.tf
+++ b/modules/workers/clusternetworks.tf
@@ -33,11 +33,7 @@ resource "oci_core_cluster_network" "workers" {
   }
 
   lifecycle {
-    ignore_changes = [
-      display_name, defined_tags, freeform_tags,
-      instance_pools["display_name"], instance_pools["defined_tags"], instance_pools["freeform_tags"],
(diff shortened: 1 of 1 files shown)


modules/workers/clusternetworks.tf
```diff
@@ -33,11 +33,7 @@ resource "oci_core_cluster_network" "workers" {
   }
 
   lifecycle {
-    ignore_changes = [
-      display_name, defined_tags, freeform_tags,
-      instance_pools["display_name"], instance_pools["defined_tags"], instance_pools["freeform_tags"],
-      placement_configuration["availability_domain"],
-    ]
+    ignore_changes = [display_name, defined_tags, freeform_tags]
```
hyder, line 36: I thought you had to put the tree in?

devoncrouse: Ignoring any key here has the effect of ignoring the whole block, unfortunately - tested a number of ways.
It doesn't look like automatic tags are being added at this level in practice.

hyder (approved):
approved

PR #666, opened 2023-03-16
feat: Support custom worker instance metadata

devoncrouse:
Adds a `worker_node_metadata` map variable for extra user-defined entries, merged with existing.


hyder (approved):
approved

PR #667, opened 2023-03-17
feat: 5.x Add configured VNICs to instance metadata when available

devoncrouse:
Makes expected VNIC attachment information available to worker nodes through instance metadata when available.


hyder:
Please provide more information about the use case/problem.

hyder (approved):
approved

PR #670, opened 2023-03-23
fix: make subnet parameters for bastion and operator optional

### LINKED ISSUE #650: var.subnets["bastion"] is required even if create_bastion_host = false
https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/8883c54278ba56302e0db54fcdd1cc5362af078a/modules/network/locals.tf#L13

hyder:
Resolves #650

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -189,11 +189,13 @@ module "network" {
   label_prefix   = var.label_prefix
 
   # oke networking parameters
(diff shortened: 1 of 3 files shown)


devoncrouse (approved):
Needs a `terraform fmt` fyi.

PR #672, opened 2023-03-23
fix: Update worker NSG global/pool handling

devoncrouse:
Minor fix for handling of worker NSG IDs between global and pool-specific values.


hyder (approved):
approved

PR #676, opened 2023-03-27
fix: Autoscaler volume-family policy, compartment rule

devoncrouse:
- Add required volume-family policy for cluster autoscaler `oci` cloud provider
- Use compartment-level policy for cluster autoscaler pending additional configuration, as `state_id` may not cover workers created in another state.


hyder:
should this fix be backported into 4.x for cluster autoscaler?

hyder (approved):
approved

PR #677, opened 2023-03-27
Update datasources.tf

robo-cap:
avoid 404 error in case the user is not tenancy admin or the cluster is not created in the root compartment.


robo-cap:
need to sign OCA

PR #678, opened 2023-03-29
feat: Add configuration for VCN gateway creation

devoncrouse:
This change maintains the current behavior by default, with new variables that allow VCN gateway creation to be disabled or forced - e.g. in cases where custom resources ([core_route_table](https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/core_route_table), [core_route_table_attachment](https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/core_route_table_attachment), etc.) are provisioned outside of the module.

```
vcn_create_nat_gateway      = "never"  # never/auto*/always
vcn_create_internet_gateway = "auto"   # never/auto*/always
vcn_create_service_gateway  = "always" # never/auto/always*
```

### DIFF
--- a/module-network.tf
+++ b/module-network.tf
@@ -28,10 +28,30 @@ module "vcn" {
   defined_tags   = var.defined_tags["vcn"]
   freeform_tags  = var.freeform_tags["vcn"]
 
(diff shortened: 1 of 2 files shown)


hyder (approved):
approved

PR #679, opened 2023-03-29
feat: 5.x Rename extension vars, provide state_id to sub-module, Multus

devoncrouse:
- Add `state_id`, `http` provider for extensions, update locals for manifest paths
- Rename extension variables to fit convention: `..._enabled` -> `..._install`
- Add Multus extension

### DIFF
--- a/variables-extensions.tf
+++ b/variables-extensions.tf
@@ -1,9 +1,9 @@
 # Copyright (c) 2017, 2023 Oracle Corporation and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
-# Calico
+# CNI: Calico
 
-variable "calico_enabled" {
+variable "calico_install" {
   default     = false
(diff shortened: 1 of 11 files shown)


variables-extensions.tf
```diff
@@ -95,7 +121,7 @@ variable "metrics_server_helm_values_files" {
 
 # Cluster autoscaler
 
-variable "cluster_autoscaler_enabled" {
+variable "cluster_autoscaler_install" {
```
hyder, line 124: I thought we supported autoscaler?

variables-extensions.tf
```diff
@@ -61,16 +61,42 @@ variable "calico_staging_dir" {
   type        = string
 }
 
+# CNI: Multus
+
+variable "multus_install" {
```
hyder, line 66: I assume we can do both and not have to pick between the 2?

variables-extensions.tf
devoncrouse: Yeah, this is just fixing the names you brought up in an earlier PR.

variables-extensions.tf
devoncrouse: Yes that's right.

hyder (approved):
Approved

PR #680, opened 2023-03-29
fix: 5.x Lifecycle ignore fixes, individual worker mode

devoncrouse:
- Fix for over-matching with node pool lifecycle ignore.
- Add `mode = "instance"` for individual self-managed nodes outside of a pool. Extended fields pending implementation aren't required with standard usage and left for context later. Provides a simple reference implementation for the `oci_core_instance` resource and can be helpful with certain testing.

### DIFF
--- a/modules/workers/instance.tf
+++ b/modules/workers/instance.tf
@@ -1,0 +1,76 @@
+resource "oci_core_instance" "workers" {
+  for_each             = local.enabled_instances
+  availability_domain  = element(each.value.availability_domains, 1)
+  compartment_id       = each.value.compartment_id
+  display_name         = each.key
+  preserve_boot_volume = false
+  shape                = each.value.shape
+
+  defined_tags  = each.value.defined_tags
+  freeform_tags = each.value.freeform_tags
+
+  dynamic "shape_config" {
+    for_each = length(regexall("Flex", each.value.shape)) > 0 ? [1] : []
+    content {
+      ocpus = each.value.ocpus
+      memory_in_gbs = ( # If > 64GB memory/core, correct input to exactly 64GB memory/core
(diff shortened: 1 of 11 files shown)


modules/workers/instance.tf
```diff
@@ -0,0 +1,134 @@
...
+  shape                = each.value.shape
+
+  defined_tags  = each.value.defined_tags
+  freeform_tags = each.value.freeform_tags
+
+  dynamic "shape_config" {
+    for_each = length(regexall("Flex", each.value.shape)) > 0 ? [1] : []
+    content {
+      ocpus = each.value.ocpus
+      memory_in_gbs = ( # If > 64GB memory/core, correct input to exactly 64GB memory/core
```
hyder, line 16: i thought we did 32/core?

modules/workers/instance.tf
```diff
@@ -0,0 +1,134 @@
...
+  lifecycle {
+    ignore_changes = [
+      defined_tags, freeform_tags, display_name,
+      metadata["cluster_ca_cert"], metadata["user_data"],
+      create_vnic_details[0].defined_tags,
+      create_vnic_details[0].freeform_tags,
+    ]
+  }
+
+  # TODO Implement configuration
```
hyder, line 82: Please remove commented code

modules/workers/instance.tf
devoncrouse: Confirmed this is consistent with [main](https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/oke/nodepools.tf#L78)

modules/workers/instance.tf
devoncrouse: Done

hyder (approved):
lgtm

PR #683, opened 2023-04-03
feat: Add oci_containerengine_cluster type variable

devoncrouse:
- Add [oci_containerengine_cluster#type](https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/containerengine_cluster#type) for [Enhanced clusters](https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengworkingwithenhancedclusters.htm)
- Update `oci` provider version to `~> 4.114.0`

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -1,4 +1,4 @@
-# Copyright (c) 2017, 2022 Oracle Corporation and/or its affiliates.
+# Copyright (c) 2017, 2023 Oracle Corporation and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 # OCI Provider parameters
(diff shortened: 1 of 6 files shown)


variables.tf
```diff
@@ -545,6 +545,16 @@ variable "cluster_name" {
   type        = string
 }
 
+variable "cluster_type" {
+  default     = "BASIC_CLUSTER"
```
hyder, line 549: use `basic` or `enhanced` for the user. Put the correct values when passing to the cluster resource itself.

hyder (changes requested):
Use `basic` or `enhanced` for the user. Put the correct values when passing to the cluster resource itself. Like we do for CNI

hyder (approved):
approved

PR #685, opened 2023-04-04
fix: Resource name-safe state_id

devoncrouse:
Use only lowercase alpha characters for generated `state_id` variable, to keep it compatible with some resource name constraints.


hyder (approved):
approved

PR #687, opened 2023-04-05
fix: 5.x Evaluate subnets definition with omitted fields correctly

devoncrouse:
Bug fix for evaluation of the `subnets` map - checking existence of key doesn't cover key present with null value.


hyder (approved):
Approved

To do: Explicit control over subnetting using newbits and netnums and not to rely on contiguous network availability.

Issue #246, opened 2020-12-04
issue with the ssh proxy jump command generated by the tf scripts


eugsim1:
Hello
I m getting this output from the tf execution :
the private key is wls-wdt-testkey-priv.txt 
Outputs:

```
bastion_public_ip = 193.123.33.61
ig_route_id = ocid1.routetable.oc1.eu-amsterdam-1.aaaaaaaafh7zup54yze7p5ddjwyyog5mkk6ailthr7weubcwlgcakew26jca
kubeconfig = export KUBECONFIG=generated/kubeconfig
nat_route_id = ocid1.routetable.oc1.eu-amsterdam-1.aaaaaaaajnrxn6m43hmxjtxr7dsg3cmglouyppwml6wwkrh5r65x7u63g54q
operator_private_ip = 10.0.1.10
ssh_to_bastion = ssh -i wls-wdt-testkey-priv.txt opc@193.123.33.61
ssh_to_operator = ssh -i wls-wdt-testkey-priv.txt -J opc@193.123.33.61 opc@10.0.1.10
```

the ssh to the operator fails 
```
 ssh -i wls-wdt-testkey-priv.txt -J opc@193.123.33.61 opc@10.0.1.10
Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
ssh_exchange_identification: Connection closed by remote host
```
The variation below works however :
```
ssh -i wls-wdt-testkey-priv.txt -o ProxyCommand="ssh -i wls-wdt-testkey-priv.txt -W %h:%p  opc@193.123.33.61"  opc@10.0.1.10
Last login: Sat Dec  5 02:16:40 2020 from bastion.bastion.oke.oraclevcn.com
```
What would be the reason that the generated ssh command with the jump flag fails in both cases i m using the same private key from the same location with the -J the ssh fails with the old syntax everything is ok !

hyder:
-J is a [simpler form of ProxyCommand](https://medium.com/maverislabs/proxyjump-the-ssh-option-you-probably-never-heard-of-2d7e41d43464) and introduced in openssh 7.3 (I think). Your ssh client might be older version and hence not supporting -J but ProxyCommand still works.

hyder:
Closing this issue. Feel free to reopen it if you think this needs further investigation

cwiggs:
[...]

PR #700, opened 2023-04-13
Fix: ssh_to_operator output use long form

cwiggs:
This commit Updates the terraform output for `ssh_to_operator` so that you can use a private key that is anywhere on the filesystem.  Currently the short form (`-J`) does not allow you to pass any arguments like `-i`, therefore if your ssh key is not in a default location that openssh knows about it will fail to connect.

More info in this issue: https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/246

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -76,7 +76,7 @@ output "operator_private_ip" {
 
 output "ssh_to_operator" {
   description = "convenient command to ssh to the operator host"
-  value       = "ssh${local.ssh_key_arg} -J ${var.bastion_user}@${local.bastion_public_ip} ${var.operator_user}@${local.operator_private_ip}"
+  value       = "ssh${local.ssh_key_arg} -o ProxyCommand='ssh ${local.ssh_key_arg} -W %h:%p ${var.bastion_user}@${local.bastion_public_ip}' ${var.operator_user}@${local.operator_private_ip}"
 }
 
 output "ssh_to_bastion" {


hyder:
Thanks for sending the PR. Can you please sign the OCA? We'll then merge

outputs.tf
```diff
@@ -76,7 +76,7 @@ output "operator_private_ip" {
 
 output "ssh_to_operator" {
   description = "convenient command to ssh to the operator host"
-  value       = "ssh${local.ssh_key_arg} -J ${var.bastion_user}@${local.bastion_public_ip} ${var.operator_user}@${local.operator_private_ip}"
+  value       = "ssh${local.ssh_key_arg} -o ProxyCommand='ssh ${local.ssh_key_arg} -W %h:%p ${var.bastion_user}@${local.bastion_public_ip}' ${var.operator_user}@${local.operator_private_ip}"
```
hyder, line 79: I think a space is needed after ssh `ssh${local.ssh_key_arg}`

hyder (changes requested):
A space is needed after the ssh command

outputs.tf
cwiggs: I left that how it was before I touched the code, it seems to output with a space so I assume `local.ssh_key_arg` has a space prefixed to it?

Either way it shouldn't effect the command if it has an extra space or not, let me know if you want me to add it and I can.

cwiggs:
> A space is needed after the ssh command

I left that how it was before I touched the code, it seems to output with a space so I assume local.ssh_key_arg has a space prefixed to it?

Either way it shouldn't effect the command if it has an extra space or not, let me know if you want me to add it and I can.

cwiggs:
Can we get this merged in?

devoncrouse (approved):

devoncrouse:
Resolves #246

outputs.tf
VizionInc: I'm the original author of this file. I can provide the local ip

PR #703, opened 2023-04-14
fix: 5.x Misc fixes, worker/deploy examples

devoncrouse:
* docs: Add Resource Manager Deploy examples
* fix: Remove unneeded subnet filter for correct creation logic
* fix: Ignore changes to display_name
* fix: Add missing cluster_type in sub-module definition
* feat: Add optional input for existing state_id
* fix: Correct self-managed policy syntax
* fix: Clean up subnets to create, consistent w/ seclists
* fix: No failure for undefined subnets when unused
* fix: Increase cluster update timeout for enhanced upgrade
* fix: Update cloud-init script
* fix: Override create map field for forced subnet
* fix: Handle null values for output
* fix: Skip tag retrieval without namespace ID

### DIFF
--- a/modules/network/subnets.tf
+++ b/modules/network/subnets.tf
@@ -61,12 +61,29 @@ locals {
     }
   }
 
+  # Create subnets if when all are true:
+  # - Associated component is enabled OR configured with create == 'always'
+  # - Subnet is configured with newbits and/or netnum/cidr
+  # - Not configured with create == 'never'
+  # - Not configured with an existing 'id'
+  subnets_to_create = { for k, v in local.subnet_info : k => v
+    if alltrue([
+      contains(keys(local.subnet_cidrs_all), k),                       # has a calculated CIDR range (not id input)
+      lookup(lookup(var.subnets, k, {}), "create", "auto") != "never", # not disabled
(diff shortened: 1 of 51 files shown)


modules/network/subnets.tf
```diff
@@ -109,18 +120,16 @@ resource "oci_core_subnet" "oke" {
...
   lifecycle {
     # TODO reflect default security_list_id instead of ignore
-    ignore_changes = [security_list_ids, freeform_tags, defined_tags, dns_label]
+    ignore_changes = [security_list_ids, freeform_tags, defined_tags, dns_label, display_name, cidr_block]
   }
 }
 
 # Create an associated security list for subnets when enabled
 # e.g. for load balancers to prevent CCM management of default security list
 resource "oci_core_security_list" "oke" {
```
hyder, line 129: So, this is to avoid the security rules going into the default security list?

hyder: Can you also add the default security list in troubleshooting mode? Sometimes, when we are experimenting, we just want to know if things will work first before going hunting for the rules.

hyder (approved):
Approved

PR #706, opened 2023-04-19
fix: Remove redundant decode for OKE init parameter

devoncrouse:
Value of `--kubelet-ca-cert` is already decoded within `oke-install.sh`.


hyder (approved):
lgtm

PR #708, opened 2023-04-22
[Draft PR] Allow vcn, cluster and nodepool in same state to be created in same run

rppala90:
Allow multiple resources ( vcn, cluster, nodepools) in same state to be created in single run.


hyder:
Can you please elaborate a bit more about the issue this is addressing?

PR #710, opened 2023-05-08
Use lower bound version specification for oci provider

12345ieee:
The oke module is a reusable module, hence it must not force the whole root module to un under a certain minor release of the oci provider.

Fixes: 830192e3e59c716f

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -6,7 +6,7 @@ terraform {
     oci = {
       source                = "oracle/oci"
       configuration_aliases = [oci.home]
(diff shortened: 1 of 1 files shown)


hyder:
Some of the new features require updated versions of the OCI provider

12345ieee:
Yes, but you need to do it with `>=`, not `~>`, or you'll effectively lock my root module.

Right now I'm forced to use 115 in the root module and can't use 116+, which would work with no issue.

hyder (approved):

PR #717, opened 2023-05-15
fix: 5.x Use correct suffix numbering for pools with mode=instance

devoncrouse:
* Include `mode: instance` pools in detailed `worker_pools` output.
* Each pool with `mode: instance` should be numbered independently rather than in a group as before.

Before:
```
poolA-0
poolA-1
poolA-2
poolB-3
poolB-4
poolB-5
```

After:
```
poolA-0
poolA-1
poolA-2
poolB-0
poolB-1
poolB-2
```

### DIFF
--- a/modules/workers/locals.tf
+++ b/modules/workers/locals.tf
@@ -156,11 +156,11 @@ locals {
   }
 
   # Enabled worker_pool map entries for individual instances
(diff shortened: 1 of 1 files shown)


rppala90 (approved):
lgtm

PR #720, opened 2023-05-17
fix: 5.x TF 1.3, subnet/NSG objects, NSG/image preconditions, unrequired VCN

### LINKED ISSUE #682: passing a list of subnet id's to OKE module fails.
It appears the bug reported in this issue https://github.com/oracle/terraform-provider-oci/issues/1662 is resolved and the attribute 
 pod_nsg_ids accepts only lists.  
 Could you please check and remove the element function in line 55. 
 
image

 
 I ran into below error when passing a list of subnet id's (string) to variable pod_ngs, the error  indicated the attribute pod_nsg_ids is of type list of strings now. 
 In my scenario I am only using the oke module https://github.com/oracle-terraform-modules/terraform-oci-oke/tree/main/modules/oke 
 
image

Could you look into this issue please. 

### Community Note

$ terraform -v 
Terraform v1.2.9
on darwin_arm64
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.2
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/oracle/oci v4.113.0

### Affected Resource(s)

oci_containerengine_node_pool

### Terraform Configuration Files

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/oke/nodepools.tf
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. 
# Please remove any sensitive information from configuration files before sharing them. 
```

### Steps to Reproduce

1. `terraform apply`

devoncrouse:
* Add preconditions for better error messaging with image/NSGs configuration issues.
* Support `create_vcn = false` and undefined `vcn_id` for workers with an existing cluster and subnets, where VCN information is not needed.
* Provide `pod_subnet_ids` as list when `var.cni_type == "npn"` - resolves #682.

Example - worker pool with existing subnet and cluster:
```
cluster_id         = "ocid1.cluster..."
create_cluster     = false
create_bastion     = false
create_nsgs        = false
create_vcn         = false
create_operator    = false
create_nsgs_always = false

worker_pools = {
  np1 = { size = 1 },
}

subnets = {
  workers = { id = "ocid1.subnet..." }
}
```

Before:
```
│ Error: Missing required argument
│ 
│   with data.oci_core_vcn.oke[0],
│   on module-network.tf line 6, in data "oci_core_vcn" "oke":
│    6:   vcn_id = var.vcn_id
│ 
│ The argument "vcn_id" is required, but no definition was found.
```

After:
```
Plan: 2 to add, 0 to change, 0 to destroy.

Changes to Outputs:
...
  + state_id               = (known after apply)
  + subnet_ids             = {
      + workers = "ocid1.subnet..."
    }
  + worker_pool_ids        = {
      + np1 = (known after apply)
    }
  + worker_subnet_id       = "ocid1.subnet..."
```

### DIFF
--- a/module-bastion.tf
+++ b/module-bastion.tf
@@ -33,12 +33,12 @@ module "bastion" {
   assign_dns          = var.assign_dns
   availability_domain = coalesce(var.bastion_availability_domain, lookup(local.ad_numbers_to_names, local.ad_numbers[0]))
   image_id            = local.bastion_image_id
-  nsg_ids             = concat(var.bastion_nsg_ids, var.create_nsgs ? [module.network.bastion_nsg_id] : [])
+  nsg_ids             = compact(flatten([var.bastion_nsg_ids, [module.network.bastion_nsg_id]]))
   is_public           = var.bastion_is_public
   shape               = var.bastion_shape
   ssh_private_key     = sensitive(local.ssh_private_key) # to await cloud-init completion
   ssh_public_key      = local.ssh_public_key
-  subnet_id           = lookup(module.network.subnet_ids, "bastion", lookup(module.network.subnet_ids, "pub_lb"))
+  subnet_id           = module.network.bastion_subnet_id
   timezone            = var.timezone
   upgrade             = var.bastion_upgrade
   user                = var.bastion_user
--- a/module-network.tf
+++ b/module-network.tf
@@ -100,8 +100,7 @@ module "network" {
   create_cluster               = var.create_cluster
   create_bastion               = var.create_bastion
   create_fss                   = var.create_fss
-  create_nsgs                  = var.create_nsgs
-  create_nsgs_always           = var.create_nsgs_always
+  nsgs                         = var.nsgs
   create_operator              = local.operator_enabled
   drg_attachments              = var.drg_attachments
   enable_waf                   = var.enable_waf
@@ -133,54 +132,56 @@ output "nat_route_table_id" {
 }
 
 # Subnets
-output "subnet_ids" {
-  description = "Map of subnet ids by role for the cluster and associated resources."
-  value       = module.network.subnet_ids
-}
-output "subnet_cidrs" {
-  description = "Map of provided/calculated subnet CIDR ranges by role for the cluster."
-  value       = module.network.subnet_cidrs
-}
 output "bastion_subnet_id" {
-  value = lookup(module.network.subnet_ids, "bastion", null)
+  value = module.network.bastion_subnet_id
+}
+output "bastion_subnet_cidr" {
+  value = module.network.bastion_subnet_cidr
 }
 output "operator_subnet_id" {
-  value = lookup(module.network.subnet_ids, "operator", null)
+  value = module.network.operator_subnet_id
+}
+output "operator_subnet_cidr" {
+  value = module.network.operator_subnet_cidr
 }
 output "control_plane_subnet_id" {
-  value = lookup(module.network.subnet_ids, "cp", null)
+  value = module.network.control_plane_subnet_id
+}
+output "control_plane_subnet_cidr" {
+  value = module.network.control_plane_subnet_cidr
 }
 output "worker_subnet_id" {
-  value = lookup(module.network.subnet_ids, "workers", null)
+  value = module.network.worker_subnet_id
+}
+output "worker_subnet_cidr" {
+  value = module.network.worker_subnet_cidr
 }
 output "pod_subnet_id" {
-  value = lookup(module.network.subnet_ids, "pods", null)
+  value = module.network.pod_subnet_id
+}
+output "pod_subnet_cidr" {
+  value = module.network.pod_subnet_cidr
 }
 output "int_lb_subnet_id" {
-  value = lookup(module.network.subnet_ids, "int_lb", null)
+  value = module.network.int_lb_subnet_id
+}
+output "int_lb_subnet_cidr" {
+  value = module.network.int_lb_subnet_cidr
 }
 output "pub_lb_subnet_id" {
-  value = lookup(module.network.subnet_ids, "pub_lb", null)
+  value = module.network.pub_lb_subnet_id
+}
+output "pub_lb_subnet_cidr" {
+  value = module.network.pub_lb_subnet_cidr
 }
 output "fss_subnet_id" {
-  value = lookup(module.network.subnet_ids, "fss", null)
+  value = module.network.fss_subnet_id
+}
+output "fss_subnet_cidr" {
+  value = module.network.fss_subnet_cidr
 }
 
 # NSGs
-output "nsg_ids" {
-  description = "Map of network security group IDs by role for the cluster and associated resources."
-  value = var.create_nsgs ? {
-    "bastion"  = module.network.bastion_nsg_id
-    "operator" = module.network.operator_nsg_id
-    "cp"       = module.network.control_plane_nsg_id
-    "int_lb"   = module.network.int_lb_nsg_id
-    "pub_lb"   = module.network.pub_lb_nsg_id
-    "workers"  = module.network.worker_nsg_id
-    "pods"     = module.network.pod_nsg_id
-    "fss"      = module.network.fss_nsg_id
-  } : null
-}
-
 output "bastion_nsg_id" {
   description = "Network Security Group for bastion host(s)."
   value       = module.network.bastion_nsg_id
--- a/module-network.tf
+++ b/module-network.tf
@@ -2,8 +2,8 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 data "oci_core_vcn" "oke" {
-  count  = var.create_vcn ? 0 : 1
-  vcn_id = var.vcn_id
+  count  = coalesce(var.vcn_id, "none") != "none" ? 0 : 1
+  vcn_id = coalesce(var.vcn_id, "none")
 }
 
 locals {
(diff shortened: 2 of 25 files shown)


module-network.tf
```diff
@@ -98,8 +100,7 @@ module "network" {
   create_cluster               = var.create_cluster
   create_bastion               = var.create_bastion
   create_fss                   = var.create_fss
-  create_nsgs                  = var.create_nsgs
-  create_nsgs_always           = var.create_nsgs_always
+  nsgs                         = var.nsgs
```
hyder, line 103: I probably asked this before but I'll ask again just to clarify:

- `never` and `always` are self-explanatory
- what does auto do?

module-network.tf
```diff
@@ -131,54 +132,56 @@ output "nat_route_table_id" {
...
 # Subnets
-output "subnet_ids" {
-  description = "Map of subnet ids by role for the cluster and associated resources."
-  value       = module.network.subnet_ids
-}
-output "subnet_cidrs" {
-  description = "Map of provided/calculated subnet CIDR ranges by role for the cluster."
-  value       = module.network.subnet_cidrs
-}
 output "bastion_subnet_id" {
```
hyder, line 135: Interesting you're splitting the outputs. Is there any particular reason?

module-network.tf
```diff
@@ -131,54 +132,56 @@ output "nat_route_table_id" {
...
-    "pub_lb"   = module.network.pub_lb_nsg_id
-    "workers"  = module.network.worker_nsg_id
-    "pods"     = module.network.pod_nsg_id
-    "fss"      = module.network.fss_nsg_id
-  } : null
+output "fss_subnet_cidr" {
+  value = module.network.fss_subnet_cidr
 }
 
+# NSGs
```
hyder, line 184: For `pod_nsg_id`, can you please clarify with the provider/service team whether this is supposed to be a single nsg or a list of nsgs? 

See: https://github.com/oracle/terraform-provider-oci/issues/1662

module-operator.tf
```diff
@@ -52,12 +52,12 @@ module "operator" {
...
   kubeconfig            = yamlencode(local.kubeconfig_private)
   kubernetes_version    = var.kubernetes_version
-  nsg_ids               = concat(var.operator_nsg_ids, var.create_nsgs ? [module.network.operator_nsg_id] : [])
+  nsg_ids               = compact(flatten([var.operator_nsg_ids, module.network.operator_nsg_id]))
   pv_transit_encryption = var.operator_pv_transit_encryption
   shape                 = var.operator_shape
   ssh_private_key       = sensitive(local.ssh_private_key) # to await cloud-init completion
   ssh_public_key        = local.ssh_public_key
-  subnet_id             = lookup(module.network.subnet_ids, "operator", "")
+  subnet_id             = module.network.operator_subnet_id
```
hyder, line 60: In the output "ssh_to_operator", do we keep the ssh command to the shorthand jump format instead of the longer one requested by #700?

modules/network/nsg-bastion.tf
```diff
@@ -2,26 +2,27 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 locals {
```
hyder, line 4: Please document the changes here to facilitate the review

modules/network/rules.tf
```diff
@@ -113,6 +113,29 @@ resource "oci_core_network_security_group_security_rule" "oke" {
       condition     = each.value.direction == "INGRESS" || coalesce(each.value.destination, "none") != "none"
       error_message = "Egress rule must have a destination: '${each.key}'"
     }
+
```
hyder, line 116: On line 91:

`
    precondition {
      condition = tostring(each.value.protocol) == tostring(local.icmp_protocol) || contains(keys(each.value), "port") || (
        contains(keys(each.value), "port_min") && contains(keys(each.value), "port_max")
      )
      error_message = "TCP/UDP rule must contain a port or port range: '${each.key}'"
    }
`
Shouldn't this be checking for `local.tcp_protocol` or `local.udp_protocol`?

variables-bastion.tf
```diff
@@ -14,8 +14,8 @@ variable "bastion_public_ip" {
 }
 
 variable "bastion_allowed_cidrs" {
-  default     = ["0.0.0.0/0"]
-  description = "A list of CIDR blocks to allow SSH access to the bastion host."
+  default     = []
+  description = "A list of CIDR blocks to allow SSH access to the bastion host. NOTE: Default is empty i.e. no access permitted. Allow access from anywhere with '0.0.0.0/0'."
```
hyder, line 18: Make sure at least a commented example is in the vars.example file

variables-cluster.tf
```diff
@@ -24,7 +24,7 @@ variable "cluster_type" {
 }
 
 variable "control_plane_is_public" {
-  default     = true
+  default     = false
   description = "Whether the Kubernetes control plane endpoint should be allocated a public IP address."
```
hyder, line 28: I think we should clarify that public here means on the public Internet

variables-cluster.tf
```diff
@@ -82,8 +82,8 @@ variable "image_signing_keys" {
 }
 
 variable "load_balancers" {
-  default     = "public"
-  description = "The type of subnets to create for load balancers."
+  default     = "internal"
```
hyder, line 85: That's not the default on OKE. If you set it to internal and a user creates a service of type load balancer without the overriding annotation, this will fail.

Please change to `both` or `public`

variables-cluster.tf
```diff
@@ -94,7 +94,7 @@ variable "load_balancers" {
 variable "preferred_load_balancer" {
   # values: public, internal.
   # When creating an internal load balancer, the internal annotation must still be specified regardless
-  default     = "public"
+  default     = "internal"
```
hyder, line 97: as above

hyder (changes requested):
Comments in line. A few interesting stuff you are doing here. I would like to understand them properly to see if we can standardize their usage in the other modules

module-network.tf
devoncrouse: >Defaults to automatic creation when ... resources are expected to utilize it.
I added notes in the variable descriptions for these to help explain - `auto` will use logic to determine if creation is deemed necessary, e.g. if operator creation is disabled, `auto` would disable the associated subnet/NSG/etc. when used. This was the standard behavior previously, and `never`/`always` provide additional control when creating only network resources/using existing ones.

Regarding the `nsgs` var change: this better aligns with the structure used for `subnets`, with more granular control over each element.

module-network.tf
devoncrouse: Yeah, some lessons learned here:
- Resource Manager works better with individual outputs - typed as ocids to automatically render links.
- Output maps can be dangerous in this context - a change detected to one entry can declare the whole output "known after apply" and cause unrelated resource recreation.
- It's more lines in this file, but individual outputs are easier to work with vs. nested lookups.

variables-cluster.tf
devoncrouse: The `internal` value aligns with `preferred_load_balancer=internal` - one of the parameters we discussed for minimized exposure by default. We can leave the existing value to match the service, though it's a bit of a departure from other changes. `both` or `public` means creating the `pub_lb` subnet/NSG by default as well.

We could also address with a note on the variable description + examples. What do you think?

module-bastion.tf
```diff
@@ -33,12 +33,12 @@ module "bastion" {
   assign_dns          = var.assign_dns
   availability_domain = coalesce(var.bastion_availability_domain, lookup(local.ad_numbers_to_names, local.ad_numbers[0]))
   image_id            = local.bastion_image_id
-  nsg_ids             = concat(var.bastion_nsg_ids, var.create_nsgs ? [module.network.bastion_nsg_id] : [])
+  nsg_ids             = compact(flatten([var.bastion_nsg_ids, [module.network.bastion_nsg_id]]))
```
devoncrouse, line 36: `var.create_nsgs` is removed; nsg_id output from network module will be null if not created, and removed from the result via `flatten`/`compact`.

module-bastion.tf
```diff
@@ -33,12 +33,12 @@ module "bastion" {
...
   is_public           = var.bastion_is_public
   shape               = var.bastion_shape
   ssh_private_key     = sensitive(local.ssh_private_key) # to await cloud-init completion
   ssh_public_key      = local.ssh_public_key
-  subnet_id           = lookup(module.network.subnet_ids, "bastion", lookup(module.network.subnet_ids, "pub_lb"))
+  subnet_id           = module.network.bastion_subnet_id
```
devoncrouse, line 41: Removed unnecessary fallback, output from network module returns provided/created accordingly.

module-network.tf
devoncrouse: Added a commit to address.

modules/network/nsg-bastion.tf
devoncrouse: Done

module-operator.tf
devoncrouse: Commit for long form is still needed in 5.x

modules/network/rules.tf
devoncrouse: This is correct: the rule gets a pass if it's icmp; otherwise it must have either `port` or (`port_min` and `port_max`).

variables-cluster.tf
devoncrouse: Done

variables-bastion.tf
devoncrouse: Done

variables-cluster.tf
devoncrouse: Retained both/public LB values as discussed to match CCM defaults.

modules/network/rules.tf
hyder: Ok, please document the code.

module-network.tf
hyder: Ok, I think we'll start taking this approach of individual outputs going forward

hyder (approved):
Approved

PR #721, opened 2023-05-19
feat: 5.x Support platform_config for preview modes

devoncrouse:
* Implement [core_instance_configuration#platform_config](https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/core_instance_configuration#platform_config) and [core_instance#platform_config](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/core_instance#platform_config).
* Update minimum Terraform version to 1.3 for [Optional Object Type Attributes with Defaults](https://www.hashicorp.com/blog/terraform-1-3-improves-extensibility-and-maintainability-of-terraform-modules).

```
  shielded_instances = {
    mode = "instance", shape = "VM.Standard2.4",
    platform_config = {
      type                               = "INTEL_VM",
      is_measured_boot_enabled           = true,
      is_secure_boot_enabled             = true,
      is_trusted_platform_module_enabled = true,
    }
  }
```


modules/workers/instanceconfig.tf
```diff
@@ -64,6 +64,23 @@ resource "oci_core_instance_configuration" "workers" {
         }
       }
 
+      dynamic "platform_config" {
+        for_each = each.value.platform_config != null ? [1] : []
+        content {
+          type                                           = lookup(each.value.platform_config, "type")
```
mehtameha1, line 70: Can we derive 'type' from 'shape' rather than exposing it as a variable? 
We can get type from data source oci_ocvp_supported_host_shapes https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/ocvp_supported_host_shapes

hyder: The ocvp data source is for OCVS, not for OKE. Does OKE API has something similar?

variables-workers.tf
```diff
@@ -230,3 +230,21 @@ variable "fss_max_fs_stat_files" {
   description = "Maximum tfiles, ffiles, and afiles values reported by NFS FSSTAT."
   type        = number
 }
+
+variable "platform_config" {
```
hyder, line 234: Please don't use object as variable type in root.

hyder (changes requested):
comments inline

modules/workers/instanceconfig.tf
mehtameha1: No idea.

variables-workers.tf
devoncrouse: As discussed, the object type isn't what it used to be with `optional` field support now - this provides better type specification for fields within, without needing to require them all as before.

modules/workers/variables.tf
```diff
@@ -48,6 +48,22 @@ variable "kubeproxy_mode" { type = string }
 variable "max_pods_per_node" { type = number }
 variable "node_labels" { type = map(string) }
 variable "node_metadata" { type = map(string) }
+variable "platform_config" {
+  default = null
+  type = object({
+    type                                           = string,
+    are_virtual_instructions_enabled               = optional(bool),
```
12345ieee, line 55: I see you switched to optional map params, nice!

If you use something like `are_virtual_instructions_enabled = optional(bool, null)`, you can get rid of the lookup in the maps above, as the value will always be present in the map, if the user doesn't give it it's the default, `null`.
It looks way less messy at the call site, but I understand it can be a stylistic choice.

See: https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes

devoncrouse: Thanks - I did end up taking advantage of this at your suggestion in the other PR for `var.nsgs` and `var.subnets`.

On further review of this case, the lookups are made against a local map rather than the input variable to honor pool-level overrides merged onto global defaults if specified - though I think it could become cleaner with `worker_pools` defined as `map(object)` in the future.

modules/workers/instanceconfig.tf
devoncrouse: It looks like [data-sources/core_shapes](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/core_shapes) has what we need:
```
            "name": "BM.GPU.B4.8",
...
            "platform_config_options": [
              {
                "access_control_service_options": [
                  {
                    "allowed_values": [
                      true,
                      false
                    ],
                    "is_default_enabled": false
                  }
                ],
                "input_output_memory_management_unit_options": [
                  {
                    "allowed_values": [
                      true,
                      false
                    ],
                    "is_default_enabled": false
                  }
                ],
                "measured_boot_options": [],
                "memory_encryption_options": [],
                "numa_nodes_per_socket_platform_options": [
                  {
                    "allowed_values": [
                      "NPS0",
                      "NPS1",
                      "NPS2",
                      "NPS4"
                    ],
                    "default_value": "NPS4"
                  }
                ],
                "percentage_of_cores_enabled_options": [],
                "secure_boot_options": [],
                "symmetric_multi_threading_options": [
                  {
                    "allowed_values": [
                      true,
                      false
                    ],
                    "is_default_enabled": true
                  }
                ],
                "trusted_platform_module_options": [],
                "type": "AMD_MILAN_BM_GPU",
                "virtual_instructions_options": [
                  {
                    "allowed_values": [
                      true,
                      false
                    ],
                    "is_default_enabled": false
                  }
                ]
              }
            ],
```

modules/workers/variables.tf
12345ieee: Oh, I'm sorry, it's hard to tell from GitHub, but I'm happy you found the info useful.

I find that this level of static typing also alleviates much of the need for documentation, so hopefully you guys get less annoyed by people in the future.

modules/workers/instanceconfig.tf
devoncrouse: Done.

examples/vars-workers-advanced.auto.tfvars
```diff
@@ -43,6 +43,16 @@ worker_pools = {
     description = "OKE-managed Node Pool with custom image", create = true,
     mode        = "node-pool", image_type = "custom", size = 1, allow_autoscaler = true,
   },
+  shielded_instances = {
+    description = "Self-managed Shielded VM Instance", create = false,
+    size        = 1, mode = "instance", shape = "VM.Standard2.4",
+    platform_config = {
+      type                               = "INTEL_VM",
```
mehtameha1, line 50: Is there a need to specify type now that it is obtained from the data source?

mehtameha1 (approved):
Looks good!

hyder (approved):
lgtm

PR #722, opened 2023-05-19
[WIP] feat: 5.x Cilium vxlan extension

devoncrouse:


hyder:
Lgtm. However, since we also have Calico, is it possible to ensure user chooses 1 of the 2?

hyder (commented):
Do we need to ensure users choose between Calico and Cilium?

hyder:
Re-implemented in #875 for simpler merge

PR #726, opened 2023-05-24
fix: 5.x Specify lifecycle ignore for extended_metadata at correct level

devoncrouse:
`extended_metadata` on the oci_core_instance_configuration resource is specified under `instance_details[0].launch_details[0].extended_metadata`.


rppala90 (approved):
lgtm

PR #727, opened 2023-05-24
fix: 5.x Merge redundant shape results for `platform_config` `type` lookup

devoncrouse:
* Resolve errors caused by multiple results for a given shape, e.g. in multi-AD regions. The fields that might differ between these aren't used for our purpose.
* Remove previously added lifecycle ignore for `extended_metadata` in favor of `{}` default or optional pool configuration, to mitigate cosmetic differences reported on subsequent apply.
```
 Error: Duplicate object key
│
│   on .terraform/modules/oci_workers/modules/workers/data-shapes.tf line 11, in locals:
│    9:   shapes_by_name = {
│   10:     for shape in data.oci_core_shapes.oke.shapes :
│   11:     lookup(shape, "name") => shape if contains(keys(shape), "name")
│   12:   }
│
│ Two different items produced the key "VM.Standard.x86.Generic" in this 'for' expression. If duplicates are expected, use the ellipsis
│ (...) after the value expression to enable grouping by key.
```

### DIFF
--- a/modules/workers/data-shapes.tf
+++ b/modules/workers/data-shapes.tf
@@ -7,12 +7,14 @@ data "oci_core_shapes" "oke" {
 
 locals {
   shapes_by_name = {
+    # Group by shape name, yielding a list of objects for each
     for shape in data.oci_core_shapes.oke.shapes :
-    lookup(shape, "name") => shape if contains(keys(shape), "name")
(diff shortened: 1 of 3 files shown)


rppala90 (approved):
lgtm

PR #738, opened 2023-06-09
fix: 5.x Add missing resources tags, materialize early

devoncrouse:
* Add resource tagging consistently, fix issues merging inputs.
* Merge common tag attributes outside of submodules to materialize earlier, i.e. visible in plan vs. "(known after apply)".
* Workers may still be known after apply due to merge with pool-level configuration, but should have tags applied correctly after.

### DIFF
--- a/variables-iam.tf
+++ b/variables-iam.tf
@@ -15,6 +15,26 @@ locals {
     ? file(var.api_private_key_path)
     : null
   )
+
+  # Merge freeform tags from map & individual inputs better suited to Resource Manager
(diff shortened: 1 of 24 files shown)


variables-iam.tf
```diff
@@ -164,6 +184,8 @@ variable "create_iam_worker_policy" {
   }
 }
 
+# Tagging
+
 variable "create_iam_tag_namespace" {
```
hyder, line 189: ideally this is is what you do in the external iam module

hyder (approved):

variables-iam.tf
devoncrouse: Agreed - I've left disabled (default) in practice and created separately for reuse. We could move out of here at some point.

PR #748, opened 2023-06-27
feat: 5.x Fill in var defs for workers submodule, deprecate FSS inputs

devoncrouse:
* Fill in variable definitions of workers submodule for better display in generated documentation
* Deprecate FSS inputs, no longer needed with CSI support
* Update worker submodule `versions.tf` for unused `home` provider, dependencies


hyder (approved):
approved

PR #779, opened 2023-07-20
fix: 5.x Unspecified default placement FDs when mode != virtual-node-pool

devoncrouse:
* Omit placement fault domains unless specified to avoid changes to existing resources, as input is not required when mode != `virtual-node-pool`.


devoncrouse:
Resolution confirmed by user.

PR #784, opened 2023-07-22
fix: fix ssh_to_operator output if create_bastion is false

### LINKED ISSUE #782: 5.x:  If Bastion is not set to be provisioned - output "ssh_to_operator" breaks.
Hello Guys,

By setting "create_bastion = false"  parameter, it brakes the deployment with a following error:
Error: Invalid template interpolation value
on .terraform/modules/oke/module-operator.tf line 104, in output "ssh_to_operator":
104:     ? "ssh${local.ssh_key_arg} -o ProxyCommand='ssh ${local.ssh_key_arg} -W %h:%p ${var.bastion_user}@${local.bastion_public_ip}' ${var.operator_user}@${local.operator_private_ip}" : null

The issue is that output do not check if create_bastion value set to true or false. If it's set to false it should not execute SSH command.
The output  in questions is this here:
https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/c3d5a1aa46795ab50b7e27566d90b5984c9e3b93/outputs.tf#L77-L80

rodrigc:
Resolves #782

Signed-off-by: Craig Rodrigues <craig@quiknode.io>

### DIFF
--- a/module-operator.tf
+++ b/module-operator.tf
@@ -100,7 +100,12 @@ output "operator_private_ip" {
 
 output "ssh_to_operator" {
   description = "SSH command for operator host"
+  #value = (local.operator_enabled || coalesce(var.operator_private_ip, "none") != "none"
+  #  ? "ssh${local.ssh_key_arg} -o ProxyCommand='ssh ${local.ssh_key_arg} -W %h:%p ${var.bastion_user}@${local.bastion_public_ip}' ${var.operator_user}@${local.operator_private_ip}" : null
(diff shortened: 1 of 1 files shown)


rodrigc:
Tested with:

```
create_operator = true
create_bastion = true
```

```
create_operator = false
create_bastion = true
```

```
create_operator = true
create_bastion = false
```

```
create_operator = false
create_bastion = false
```

hyder (approved):
Approved

Issue #768, opened 2023-07-14
OKE provisioning on sub-compartment which has no direct policy permission at the tenancy level


thpham:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The aim of this request is to be able to use this terraform module under a sub-compartment hierarchy which the provisioner has the full admin policies, but not at the root tenancy/compartment level.
Moreover, we should be able to leverage usage of an alternative identity domain for dynamic-group and service account users that are at the top level of this sub-compartment hierarchy.

### New or Affected Resource(s)

I'm targeting the `5.x` development branch.

affected terraform inner-modules are:

- iam

### Potential Terraform Configuration

```hcl
# Copy-paste any Terraform configurations for how the requested feature may be used. 
```

PR #785, opened 2023-07-24
5.x: OKE provisioning on sub-compartment with support of custom identity domain.

thpham:
This PR relates to the issue #768.

OKE provisioning on sub-compartment which has no direct policy permission at the tenancy level.
This also support the case which has a custom identity domain in the sub-compartment instead of using the Default one at the tenancy level.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -8,6 +8,7 @@
   - [Identity](./guide/identity.md)
     - [Policies](./guide/identity_policies.md)
     - [Tags](./guide/identity_tags.md)
+    - [Identity Domain](./guide/identity_domain.md)
   - [Network](./guide/network.md)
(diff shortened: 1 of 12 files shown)


hyder:
Hi @thpham, 

Thank you for your contribution. Can you also please add some documentation around usage? You can stick under identity: https://oracle-terraform-modules.github.io/terraform-oci-oke/guide/identity.html, we'll reorganize if necessary.

thpham:
Hi @hyder,

thank you for your feedback ! I added some infos about the config and use case-. Would it be of for you ?

thpham:
@devoncrouse , what should I do with this PR, are you considering it and therefore I take the time to update it towards the merged main branch ? or you simply don't think it is interesting and therefore I close it and take another approach (fork)

PR #796, opened 2023-08-13
Fix: Ignore prohibit_public_ip_on_vnic changes in control plane subnet

SG60:
If `prohibit_public_ip_on_vnic` isn't ignored, then changing the k8s API endpoint between public and private requires recreating the entire cluster.

### DIFF
--- a/modules/network/subnets.tf
+++ b/modules/network/subnets.tf
@@ -12,7 +12,7 @@ resource "oci_core_subnet" "cp" {
   vcn_id                     = var.vcn_id
 
   lifecycle {
-    ignore_changes = [dns_label]
+    ignore_changes = [dns_label, prohibit_public_ip_on_vnic]
   }
 }
 


12345ieee:
Ah, it comes full circle.

For anyone confused why this is relevant now, I asked for the behaviour to be corrected in the provider here: https://github.com/oracle/terraform-provider-oci/issues/1834 , before it used to give absurd plans that would never go in.

In a better world the control plane subnet would be updatable by accepting some downtime, but this PR is definitely 2nd best.

hyder (approved):
Approved.

hyder:
@devoncrouse I've tested switching from public to private and back to public. This works great. Can you add this to the 5.x branch as well?

hyder:
@SG60 thank you for your contribution.

SG60:
@hyder thanks for merging!

PR #800, opened 2023-08-15
fix: 5.x Update Cluster Autoscaler docs, image tag

devoncrouse:
- Update Cluster Autoscaler docs/examples with correct `min_size`/`max_size` parameters
- Update Cluster Autoscaler deployment image tag version -> `1.26.2-7` ([reference](https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengusingclusterautoscaler.htm))


hyder (approved):
approved

PR #801, opened 2023-08-15
fix: 5.x Evaluation for auto NSG creation with defaults

devoncrouse:
The fallback to "auto" was not working properly as a value of `null` was found:
```
lookup(local.bastion_nsg_config, "create", "auto") == "always",
```

This is now coalesced to the default value "auto" as expected.


hyder (approved):
approved

PR #804, opened 2023-08-17
fix: missing lb rule doesn't allow explicit ingress from anywhere

hyder:


modules/network/rules.tf
```diff
@@ -133,6 +133,16 @@ resource "oci_core_network_security_group_security_rule" "oke" {
...
           contains(var.control_plane_allowed_cidrs, local.anywhere),
         ]),
+
+        # TCP ingress to internal load balancer from anywhere has been configured explicitly
+        alltrue([
+          contains(keys(var.allow_rules_internal_lb), each.key),
+        ]),
+
+        # TCP ingress to public load balancer from anywhere has been configured explicitly
+        alltrue([
```
devoncrouse, line 143: Redundant, remove alltrue

devoncrouse (approved):

PR #852, opened 2023-10-19
fix: VCN 3.5.4 -> 3.5.3 due to plan changes

devoncrouse:


hyder:
This causing a cyclic error as discussed. I think we should address the problem in the VCN module first.

PR #861, opened 2023-10-29
fix tiny error in docs depicting topology

satyamkapoor:
Tiny mistake in documentation of topology where in an image's caption, instead of Deploying private workers, it is incorrectly written deploying public workers.


satyamkapoor:
OCA under review

hyder (approved):

hyder:
Well spotted. thank you

PR #872, opened 2023-11-20
fix: Remove sensitive mask

abatilo:
This value is an HTTP response to a publicly accessible daemonset
manifest but it's being marked as sensitive.

It's marked as sensitive, but then it's used in the body of an error
message for a `precondition`:
https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/extensions/rdma_cni_plugin.tf#L61

Because of this variable being in the error message, `terraform plan`
will exit with exit code `1` even though the plan itself actually
succeeded.

This variable shouldn't be marked sensitive anyways, as its holding the
value for an already public yaml file.

PR #873, opened 2023-11-22
fix: Remove sensitive mask from RDMA CNI

OguzPastirmaci:
Duplicate of #872 with the OCA accepted.

---

This value is an HTTP response to a publicly accessible daemonset
manifest but it's being marked as sensitive.

It's marked as sensitive, but then it's used in the body of an error
message for a precondition:
https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/extensions/rdma_cni_plugin.tf#L61

Because of this variable being in the error message, terraform plan
will exit with exit code 1 even though the plan itself actually
succeeded.

This variable shouldn't be marked sensitive anyways, as its holding the
value for an already public yaml file.


hyder (approved):
approved. thank you for your contribution

hyder:
@OguzPastirmaci we are including your contribution into main by copying it directly and we'll close the PR here.

Issue #514, opened 2022-04-21
Refer to LPGs created by the module in the `nat/internet_gateway_route_rules`


12345ieee:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

#295 (resolved in #332 ) added the capability to create LPGs in the oke VCN by the use of:
```hcl
local_peering_gateways  = {lpg1_name = { route_table_id = "", peer_id = "possible_id_of_peered" }}
```
But it's not possible to refer to these gateways in the `nat/internet_gateway_route_rules` sections the way it's possible for the predefined `drg`, `nat_gateway`, `internet_gateway`.

This makes it impossible to create fully contained LPG-route networks in the OKE module, as it holds the VCN inside, so one has a chicken-and-egg problem.
#467 would solve this from another angle, but the support needed is SO SO close to what we already have.

### New or Affected Resource(s)

```
oci_core_local_peering_gateway
oci_core_route_table
```
### Potential Terraform Configuration

I suggest a configuration similar to the one for `drg`, `nat_gateway`, `internet_gateway`:
```hcl
  local_peering_gateways  = {lpg1_name = { route_table_id = "", peer_id = "possible_id_of_peered" }}
  nat_gateway_route_rules = [
    {
      destination       = "172.18.0.0/16",
      destination_type  = "CIDR_BLOCK",
      network_entity_id = "nat_gateway",
      description       = "already works, uses the id of the NAT GW created by the OKE module"
    },
    {
      destination       = "172.19.0.0/16",
      destination_type  = "CIDR_BLOCK",
      network_entity_id = "LPG@lpg1_name",
      description       = "this doesn't work right now, the network_entity_id would be the id of the LPG created above"
    }
  ]
```

hyder:
Hi @12345ieee, 

For the `network_entity_id` value, can you please try passing the LPGs' OCIDs directly?

I would also like to understand your use case better. Can you please elaborate?

12345ieee:
Hi @hyder ,

I would like to put the LPG ocid there, the issue is that I do not have it until the module runs, because the LPG is created by the module.
[...]

PR #874, opened 2023-11-23
feat: networking module - bump  to v3.6.0

thpham:
This PR is bumping the internal dependency on `terraform-oci-vcn` to its latest version `v3.6.0` which [integrates the following changes](https://github.com/oracle-terraform-modules/terraform-oci-vcn/compare/v3.5.5...v3.6.0).

This also resolves issue #514.


hyder:
Would it be possible for you to sign the OCA? It will make collaborating with us so much faster.

thpham:
> Would it be possible for you to sign the OCA? It will make collaborating with us so much faster.

I already signed the OCA, proof is that I already created multiple PR [here](https://github.com/oracle-terraform-modules/terraform-oci-oke/pull/785) (example) and in other terraform modules. I think there is a problem with your bot.

hyder:
I thought so. I'll check internally.

hyder (approved):
Approved. Thank you for your patience

PR #880, opened 2023-12-07
feat: upgraded default Autonomous Linux to 8.8

hyder:
disabled unnecessary repos

### DIFF
--- a/modules/bastion/cloudinit.tf
+++ b/modules/bastion/cloudinit.tf
@@ -8,6 +8,15 @@ data "cloudinit_config" "bastion" {
 
   part {
     content_type = "text/cloud-config"
+    # https://cloudinit.readthedocs.io/en/latest/reference/examples.html#run-commands-on-first-boot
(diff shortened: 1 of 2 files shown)


devoncrouse (approved):

modules/bastion/cloudinit.tf
```diff
@@ -6,6 +6,15 @@ data "cloudinit_config" "bastion" {
   gzip          = true
   base64_encode = true
 
+  part {
+    content_type = "text/cloud-config"
+    # https://cloudinit.readthedocs.io/en/latest/reference/examples.html#run-commands-on-first-boot
+    content = <<-EOT
+    runcmd:
+    - dnf config-manager --disable ol8_addons --disable ol8_appstream
```
Djelibeybi, line 14: I wouldn't recommend disabling `ol8_appstream` as almost all installs will have RPMs from that repo. If you disable it, they will never be updated, i.e. you could be missing security fixes.

Issue #883, opened 2024-01-02
Terraform plan/apply fail with invalid count/for_each if compartment or VCN created in same configuration but outside module


chrizkim:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version and Provider Version

$ terraform -v
Terraform v1.6.5 on darwin_amd64

### Affected Resource(s)

* `data.oci_core_vcn.oke`
* `module.iam.data.oci_identity_tags.oke`
* `module.network.oci_core_subnet.oke`
* `module.network.oci_core_security_list.oke`
* `data.oci_identity_availability_domains.all`
* `module.workers.data.oci_identity_fault_domains.all`

### Terraform Configuration Files

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/module-network.tf#L4-L5

```hcl
data "oci_core_vcn" "oke" {
  count  = coalesce(var.vcn_id, "none") != "none" ? 1 : 0
...
```

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/iam/tagging.tf#L16-L17

```hcl
data "oci_identity_tags" "oke" {
  count            = var.create_iam_resources && local.tag_namespace_id_found != null ? 1 : 0
...
```

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/network/subnets.tf#L147-L151

```hcl
resource "oci_core_security_list" "oke" {
  for_each = {
    for k, v in local.subnets_to_create : k => v
    if tobool(lookup(v, "create_seclist", false))
  }
...
```

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/modules/workers/data-faultdomains.tf#L4-L5

```hcl
data "oci_identity_fault_domains" "all" {
  for_each            = var.ad_numbers_to_names
...
```

### Debug Output

```
╷
│ Error: Invalid count argument
│
│   on .terraform/modules/oke/module-network.tf line 5, in data "oci_core_vcn" "oke":
│    5:   count  = coalesce(var.vcn_id, "none") != "none" ? 1 : 0
│
[...]

PR #884, opened 2024-01-02
fix: Remove unknown resource counts from derived inputs

chrizkim:
Terraform plan and apply throw errors if a Compartment is created in the same configuration using this module and passed in through `var.compartment_id`. Same for VCNs created in the same configuration but not created by this module directly (i.e. passed in through `var.vcn_id` with `var.create_vcn = false`)

Fixes for #883

### DIFF
--- a/module-iam.tf
+++ b/module-iam.tf
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 data "oci_identity_availability_domains" "all" {
-  compartment_id = local.compartment_id
+  compartment_id = local.tenancy_id != "unknown" ? local.tenancy_id : local.compartment_id
 }
 
 locals {
--- a/module-network.tf
+++ b/module-network.tf
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 data "oci_core_vcn" "oke" {
-  count  = coalesce(var.vcn_id, "none") != "none" ? 1 : 0
+  count  = var.create_vcn ? 0 : 1
   vcn_id = coalesce(var.vcn_id, "none")
 }
 
--- a/modules/iam/tagging.tf
+++ b/modules/iam/tagging.tf
@@ -13,13 +13,6 @@ data "oci_identity_tag_namespaces" "oke" {
   state = "ACTIVE" // TODO Support reactivation of retired namespace w/ update
 }
 
-data "oci_identity_tags" "oke" {
-  count            = var.create_iam_resources && local.tag_namespace_id_found != null ? 1 : 0
-  provider         = oci.home
-  tag_namespace_id = local.tag_namespace_id_found
-  state            = "ACTIVE" // TODO Support reactivation of retired tag w/ update
-}
-
 locals {
   # Filtered value from data source (only 1 by name, or null)
   # Identified tag namespace ID when not created and used
@@ -31,7 +24,6 @@ locals {
     var.create_iam_resources,
     var.create_iam_tag_namespace,
     local.tag_namespace_id_found == null,
-    one(data.oci_identity_tags.oke[*].tags) == null,
   ])
 
   # Map of standard tags & descriptions to be created if enabled
@@ -64,17 +56,17 @@ resource "oci_identity_tag_namespace" "oke" {
   description    = "Tag namespace for OKE resources"
   name           = var.tag_namespace
   # defined_tags   = local.defined_tags
-  freeform_tags  = local.freeform_tags
+  freeform_tags = local.freeform_tags
   lifecycle {
     ignore_changes = [defined_tags, freeform_tags]
   }
 }
 
 resource "oci_identity_tag" "oke" {
-  provider         = oci.home
-  for_each         = local.create_iam_tag_namespace ? local.tags : {} #{ for k, v in oci_identity_tag_namespace.oke : k => local.tags } # local.create_iam_tag_namespace ? local.tags : {}
-  description      = each.value
-  name             = each.key
+  provider    = oci.home
+  for_each    = local.create_iam_tag_namespace ? local.tags : {} #{ for k, v in oci_identity_tag_namespace.oke : k => local.tags } # local.create_iam_tag_namespace ? local.tags : {}
+  description = each.value
+  name        = each.key
   # defined_tags     = local.defined_tags
   freeform_tags    = local.freeform_tags
   tag_namespace_id = one(oci_identity_tag_namespace.oke[*].id)
--- a/modules/network/subnets.tf
+++ b/modules/network/subnets.tf
@@ -79,7 +79,7 @@ locals {
   # - Subnet is configured with newbits and/or netnum/cidr
   # - Not configured with create == 'never'
   # - Not configured with an existing 'id'
-  subnets_to_create = length(var.vcn_cidrs) > 0 ? merge(
+  subnets_to_create = merge(
     { for k, v in local.subnet_info : k =>
       # Override `create = true` if configured with "always"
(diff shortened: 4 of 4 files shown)


module-network.tf
```diff
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 data "oci_core_vcn" "oke" {
-  count  = coalesce(var.vcn_id, "none") != "none" ? 1 : 0
+  count  = var.create_vcn ? 0 : 1
```
chrizkim, line 5: This data source is only used here: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/main/module-network.tf#L14-L16

As it currently works, if `var.create_vcn == true` this gets ignored entirely, and if `var.create_vcn == false` and `var.vcn_id` is empty, then this module fails elsewhere already

modules/iam/tagging.tf
```diff
@@ -13,13 +13,6 @@ data "oci_identity_tag_namespaces" "oke" {
   state = "ACTIVE" // TODO Support reactivation of retired namespace w/ update
 }
 
-data "oci_identity_tags" "oke" {
-  count            = var.create_iam_resources && local.tag_namespace_id_found != null ? 1 : 0
-  provider         = oci.home
-  tag_namespace_id = local.tag_namespace_id_found
-  state            = "ACTIVE" // TODO Support reactivation of retired tag w/ update
-}
-
```
chrizkim, line 22: Only used on line 34 (which I removed)

modules/iam/tagging.tf
```diff
@@ -31,7 +24,6 @@ locals {
     var.create_iam_resources,
     var.create_iam_tag_namespace,
     local.tag_namespace_id_found == null,
-    one(data.oci_identity_tags.oke[*].tags) == null,
```
chrizkim, line 34: If the tag namespace id is found, then this looks for if the namespace has any tags in it, but the previous line/boolean already ensures that the tag namespace id should not be found, so this line never actually matters

modules/network/subnets.tf
```diff
@@ -79,7 +79,7 @@ locals {
   # - Subnet is configured with newbits and/or netnum/cidr
   # - Not configured with create == 'never'
   # - Not configured with an existing 'id'
-  subnets_to_create = length(var.vcn_cidrs) > 0 ? merge(
+  subnets_to_create = merge(
```
chrizkim, line 82: No reason to safeguard against an empty `var.vcn_cidrs` list, the plan will already fail elsewhere if `length(var.vcn_cidrs) == 0`

module-iam.tf
```diff
@@ -2,7 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 data "oci_identity_availability_domains" "all" {
-  compartment_id = local.compartment_id
+  compartment_id = local.tenancy_id != "unknown" ? local.tenancy_id : local.compartment_id
```
chrizkim, line 5: This one is more a limitation of the OCI API and the Terraform provider. If the `compartment_id` is not known during planning, then Terraform does not know how many `data.oci_identity_availability_domains.all.availability_domains` will exist. When eventually used as a `for_each` for `module.workers.data.oci_identity_fault_domains.all`, it will fail.

Best option is to try to use the static Tenancy OCID if possible and only fallback to the Compartment OCID if that is unknown

modules/network/subnets.tf
devoncrouse: Not sure if it's the case here, but some defensive coding in the project has related to errors on destroy, such as missing inputs that shouldn't matter blocking cleanup of known state, and we may need to handle the error vs. fail elsewhere.

chrizkim: @devoncrouse Thanks for looking through this PR. Understand the potential issue, any idea of a potential error case I can try to reproduce to see what workarounds are possible? Without a bit more context, one failsafe i can think of is to replace
```
subnets_to_create = length(var.vcn_cidrs) > 0 ? merge(...) : {}
```
with 
```
subnets_to_create = try(merge(...), {})
```
(assuming the issue is that calculating `subnets_to_create` throws an error when `var.vcn_cidrs` is empty).
This avoids `local.subnets_to_create` from having a dependency on `data.oci_core_vcn.oke` from the root module, which is where this problem comes from.

Let me know thoughts, thanks!

devoncrouse: @chrizkim :+1: that seems safe, thanks!

module-iam.tf
devoncrouse: The one issue here is for some scenarios where auth != resource tenancy, and the returned AD prefixes are incorrect when the same provider value is used.

Another workaround I can think of might be to define a map local with 3 static entries each conditionally set if available from the data source, so the count is known?

chrizkim: @devoncrouse Is the scenario you described that if the provider auth does not have access to the whole tenancy and is restricted to a subcompartment, then listing the availability domains against the tenancy OCID fails? Or do you mean there is a scenario where the subcompartment's ADs differ from the parent tenancy's ADs?

Regarding the local map, if I understood correctly, I still don't think it fully resolves the dependency. If the value of the local map depends on a data source, and that map gets used in a `for_each` or `count`, Terraform will still be relying on the data source to know the count. Only similar approach I can see is just hard-coding `FD-1/2/3` and not bothering with `data.oci_identity_fault_domains` at all. Let me know if I'm missing something

modules/network/subnets.tf
chrizkim: done

hyder:
hi @chrizkim, can you please sign the contributor agreement so we can merge when this review is complete?

chrizkim:
@hyder Signed one a couple weeks back but I just checked and it seems that it got rejected with no note as to why. I'll ask our Oracle reps to look into what happened

hyder:
> @hyder Signed one a couple weeks back but I just checked and it seems that it got rejected with no note as to why. I'll ask our Oracle reps to look into what happened

Thanks for letting us know @chrizkim.

@alina-yur Is there anything we can do on our side to help?

alina-yur:
Hi @hyder. There's an email thread about this issue; the problem is in missing required information in the OCA request. Once we receive that information, we can approve the PR and the check will pass.

modules/network/subnets.tf
robo-cap: This still fails if, in the `subnets `variable, we are providing the `newbits `instead of the `cidr` attributes.

Here is an example of the subnets variable that can be used for test:

```
    subnets = {
        bastion  = { newbits = 13 }
        operator = { newbits = 13 }
        cp       = { newbits = 13 }
        int_lb   = { newbits = 11 }
        pub_lb   = { newbits = 11 }
        workers  = { newbits = 2 }
        pods     = { newbits = 2 }
        # bastion  = { cidr = "10.0.0.0/29" }
        # operator = { cidr = "10.0.0.8/29" }
        # cp       = { cidr = "10.0.0.16/29" }
        # int_lb   = { cidr = "10.0.0.64/27" }
        # pub_lb   = { cidr = "10.0.0.128/27" }
        # workers  = { cidr = "10.0.8.0/21" }
    }
```


To work around the problem with `newbits` is necessary to remove one validation from the `subnets_to_create` variable:

`# contains(keys(local.subnet_cidrs_all), k),                                              # has a calculated CIDR range (not id input)`

module-iam.tf
robo-cap: @devoncrouse to cover the use case you described we should be more specific in the example we provide for the provider authentication.

`tenancy_id`/`tenancy_ocid` should be used to target the resource tenancy.
`auth_tenancy_id` should be used to override the above in the provider configuration.

Most of the users of this module consider `tenancy_id` as the resource tenancy and for me, it makes sense to use it for ADs name.

bobturneruk:
Hi! I appreciate the work that's been done here. Is the plan that this PR will be merged, please? And if so, when will that happen?

hyder (approved):
approved

Issue #869, opened 2023-11-14
Wrong `control_plane_is_public` behavior for OKE cluster


Noksa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

```
resource "oci_containerengine_cluster" "k8s_cluster"
resource "oci_core_subnet" "oke"
resource "oci_containerengine_node_pool" "workers"
```

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. 
# Please remove any sensitive information from configuration files before sharing them. 
```

### Expected Behavior

I think `control_place_is_public` should work the same way as it works via `web` console.
We should be able to enable/disable public endpoint at any time changing `control_place_is_public` if we create it with `true` (cp subnet is public).

For instance, if we create `OKE` cluster using `web` console in public subnet, we are able to enable/disable public api endpoint at any time:
<img width="1298" alt="image" src="https://github.com/oracle-terraform-modules/terraform-oci-oke/assets/36312857/98942764-6c1c-48dd-8e0d-dc45d563c4ce">

### Actual Behavior

Currently, it is impossible to change `control_place_is_public` from `true` to `false` without recreating the whole OKE cluster.

Using terraform, if we created an OKE cluster with `control_plane_is_public` set to `true` and we change it to `false`, terraform wants to recreate the whole OKE cluster from scratch. I guess the root cause might be here: https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/667bb239fcdc57da59fa74e64c448edb4529b34c/modules/network/subnets.tf#L131

When it is `false`, public subnet wants to be private instead because of `prohibit_public_ip_on_vnic`

### Steps to Reproduce

Use 

```
module "oke" {
  source  = "oracle-terraform-modules/oke/oci"
  version = "5.0.0"
}
```

1. Just create a simple cluster using terraform with `control_plane_is_public` set to `true`
2. Change `control_plane_is_public` to false
3. Run `terraform plan` and check that everything should be replaced

```
[...]

PR #888, opened 2024-01-25
Fix : Wrong control_plane_is_public behavior for OKE cluster 

syedthameem85:
Resolves[https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/869]

@Noksa / @hyder / @devoncrouse
With public subnets, we have option to either assign public IP or defer it for assignment at a later point of time. This is the feature that is being used to assign or unassign the public IP to Control plane API endpoint in OCI console. So if the cluster has been created with public subnet with/without actually assigning the public IP at the time off creation, there should be an option to toggle the public IP after cluster is provisioned.

The variable control_plane_is_public is being used in code to decide the subnet type (public or private). If we toggle this variable, the subnet recreation will kick off which will recreate the cluster and nodepools.

For solution to toggle the public ip , we need to introduce a new boolean variable (assign_public_ip_to_control_plane) to toggle public ip on control plan API endpoint.

With control_plane_is_public set to true, assign_public_ip_to_control_plane can be set to either true or false.

When assign_public_ip_to_control_plane is set to true, the public IP will be assigned to Control plane API endpoint.
When assign_public_ip_to_control_plane set to false, although the subnet is public, the public IP will not be assigned to the API endpoint.
With control_plane_is_public set to false, public ip will not be allocated to API endpoint irrespective of value set for assign_public_ip_to_control_plane

### DIFF
--- a/modules/cluster/cluster.tf
+++ b/modules/cluster/cluster.tf
@@ -16,7 +16,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   }
 
   endpoint_config {
-    is_public_ip_enabled = var.control_plane_is_public ? var.assign_public_ip_to_control_plane : var.control_plane_is_public
+    is_public_ip_enabled = var.control_plane_is_public && var.assign_public_ip_to_control_planeßß
     nsg_ids              = var.control_plane_nsg_ids
     subnet_id            = var.control_plane_subnet_id
   }
--- a/modules/cluster/cluster.tf
+++ b/modules/cluster/cluster.tf
@@ -16,7 +16,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   }
 
   endpoint_config {
-    is_public_ip_enabled = var.control_plane_is_public
+    is_public_ip_enabled = var.control_plane_is_public ? var.assign_public_ip_to_control_plane : var.control_plane_is_public
     nsg_ids              = var.control_plane_nsg_ids
     subnet_id            = var.control_plane_subnet_id
   }
--- a/modules/cluster/cluster.tf
+++ b/modules/cluster/cluster.tf
@@ -16,7 +16,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   }
 
   endpoint_config {
-    is_public_ip_enabled = var.control_plane_is_public && var.assign_public_ip_to_control_planeßß
+    is_public_ip_enabled = var.control_plane_is_public && var.assign_public_ip_to_control_plane
     nsg_ids              = var.control_plane_nsg_ids
     subnet_id            = var.control_plane_subnet_id
   }
--- a/variables-cluster.tf
+++ b/variables-cluster.tf
@@ -29,6 +29,12 @@ variable "control_plane_is_public" {
   type        = bool
 }
 
+variable "assign_public_ip_to_control_plane" {
+  default     = false
+  description = "An additional variable to toggle public IP on Kubernetes control plane endpoint when its public."
+  type        = bool
+}
+
 variable "control_plane_nsg_ids" {
   default     = []
   description = "An additional list of network security groups (NSG) ids for the cluster endpoint."
--- a/variables-cluster.tf
+++ b/variables-cluster.tf
@@ -31,7 +31,7 @@ variable "control_plane_is_public" {
 
 variable "assign_public_ip_to_control_plane" {
   default     = false
-  description = "An additional variable to toggle public IP on Kubernetes control plane endpoint when its public."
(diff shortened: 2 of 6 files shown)


hyder:
What is the subnet visibility and behaviour when doing this swap in the OCI console?

syedthameem85:
> What is the subnet visibility and behaviour when doing this swap in the OCI console?

@hyder - The public IP toggle IP option is available for Kubernetes API Endpoint subnet. The OCI console forces user to provide public subnet if public IP assignment is checked.

hyder:
Why not make the control plane subnet public anyway and then use the existing variable to decide whether to allocate a public IP or not?

syedthameem85:
> Why not make the control plane subnet public anyway and then use the existing variable to decide whether to allocate a public IP or not?

This cannot be achieved with one variable. If you are saying that force the control plane subnet to be public always, this would not be inline with console. When you set control_plane_is_public to true, how would we decide the case where subnet is public while public IP is not required?

hyder:
A public subnet can have both public and private IP addresses. 

If the idea is be able to switch between a publicly accessible control plane and a private one, then you use the existing variable and keep the subnet public. Otherwise, this code is redundant:

```
is_public_ip_enabled = var.control_plane_is_public ? var.assign_public_ip_to_control_plane : var.control_plane_is_public
```

syedthameem85:
@hyder - We have 2 conditions that need to be addressed here - first condition is whether the subnet is public or private and when subnet is public, whether we need to assign public ip or not. To serve these 2 conditions, we need to have 2 variables.

hyder:
> @hyder - We have 2 conditions that need to be addressed here - first condition is whether the subnet is public or private and when subnet is public, whether we need to assign public IP or not. To serve these 2 conditions, we need to have 2 variables.

I don't see the need for 2 variables here. The only reason you want to make the subnet public is to have a public IP address. Since a public subnet can have both public and private IPs and the intent is to be able to toggle the visibility of the control plane, then use the existing variable to toggle the visibility instead of the subnet. 

What is the impact on routing table, use of DRG for private connectivity to the control plane?

syedthameem85:
This code is inline with OCI console behavior. When someone toggles the Public IP for API endpoint in OCI console, its doesn't go ahead and recreate the subnet (and it should not do that). The subnet still stays public and its just the IP that is toggled. So, there shouldn't be any changes required at network level (be it routing table or DRG).

variables-cluster.tf
```diff
@@ -29,6 +29,12 @@ variable "control_plane_is_public" {
   type        = bool
 }
 
+variable "assign_public_ip_to_control_plane" {
+  default     = false
+  description = "An additional variable to toggle public IP on Kubernetes control plane endpoint when its public."
```
hyder, line 34: Change description to: 

"Whether to assign a public IP address to the API endpoint for public access. Requires the control plane subnet to be public to assign a public IP address."

modules/cluster/cluster.tf
```diff
@@ -16,7 +16,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   }
 
   endpoint_config {
-    is_public_ip_enabled = var.control_plane_is_public
+    is_public_ip_enabled = var.control_plane_is_public ? var.assign_public_ip_to_control_plane : var.control_plane_is_public
```
hyder, line 19: you need to review the logic on this. it should be if both control_plane and assign_public_ip are true, then true, else false.

variables-cluster.tf
syedthameem85: @hyder -Fixed

modules/cluster/cluster.tf
syedthameem85: @hyder -Fixed

syedthameem85:
> Please also add the variable in the vars example for cluster and in the documentation

@hyder - Done

modules/cluster/cluster.tf
```diff
@@ -16,7 +16,7 @@ resource "oci_containerengine_cluster" "k8s_cluster" {
   }
 
   endpoint_config {
-    is_public_ip_enabled = var.control_plane_is_public
+    is_public_ip_enabled = var.control_plane_is_public && var.assign_public_ip_to_control_planeßß
```
hyder, line 19: you have extra characters at the end of this line.

hyder (changes requested):
please remove extra characters

modules/cluster/cluster.tf
syedthameem85: @hyder - Done

hyder (approved):
approved

PR #890, opened 2024-02-05
Improve operator pkg installation

robo-cap:
- upgrade to operator olcne reppo to version 1.8 
- ability to install kubectl using curl from [k8s.io](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-on-linux)

### DIFF
--- a/modules/operator/cloudinit.tf
+++ b/modules/operator/cloudinit.tf
@@ -103,7 +103,7 @@ data "cloudinit_config" "operator" {
 
   # kubectl installation
   dynamic "part" {
-    for_each = var.install_kubectl_from_repo ? [0] : [1]
+    for_each = var.install_kubectl_from_repo ? [] : [1]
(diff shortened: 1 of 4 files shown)


modules/operator/cloudinit.tf
```diff
@@ -101,6 +101,24 @@ data "cloudinit_config" "operator" {
     merge_type = local.default_cloud_init_merge_type
   }
 
+  # kubectl installation
+  dynamic "part" {
+    for_each = var.install_kubectl_from_repo ? [0] : [1]
```
devoncrouse, line 106:
```suggestion
    for_each = var.install_kubectl_from_repo ? [1] : []
```
Looks like the element in for_each clauses will have this run in both cases vs. empty else

devoncrouse (approved):

PR #896, opened 2024-02-08
feat: Configurable boot volume VPUs/GB on self-managed

devoncrouse:
- Support `boot_volume_vpus_per_gb` in `worker_shape` and entries of `worker_pools` for self-managed modes


hyder (approved):
When I saw the title, I was wondering how you would ensure only BYON workers would be supported. Turns out with the new worker node approach, we already can! Approved!

PR #909, opened 2024-03-28
Improve logic for kube_config datasource

robo-cap:
Fixes #908 .

PR #910, opened 2024-03-28
fix issue introduced by #909 with new clusters and #873

robo-cap:
Fix issues introduced by #909 for new clusters.
Add the changes in #873.


hyder (approved):
approved

PR #912, opened 2024-04-03
Update versions.tf

livferda:


devoncrouse:
@livferda this is no longer maintained under hashicorp, the current source is correct

PR #927, opened 2024-05-28
Allow ignore changes for node pools

VladislavGeneDx:


hyder:
Hi,

Thanks for your contribution. Can you please provide a description of your changes and the use case behind?

robo-cap:
I agree that we should ignore changes to the size of the nodepool on the cluster-autoscaler-managed nodepools.

It's difficult to make this code look good now considering the limitations on terraform:
- https://github.com/hashicorp/terraform/issues/33236
- https://github.com/hashicorp/terraform/issues/24188

PR #936, opened 2024-06-28
Add support to ignore_size attribute on nodepools

### LINKED ISSUE #763: Add option to ignore changes to node pool size
### Community Note

### Description

The oci_containerengine_node_pool resource currently has the ignore_changes lifecycle rule applied on [various attributes](https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/c3d5a1aa46795ab50b7e27566d90b5984c9e3b93/modules/oke/nodepools.tf#L136) but this does not include node_config_details.size. 

This presents an issue when using the cluster autoscaler because it results in Terraform trying to revert the node count in the pool to the value specified by this variable. This behaviour may not be desired if the autoscaler has scaled up/down since the last Terraform apply operation and you want the nodes currently in place to remain unaffected following future apply operations.

The [documentation](https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengusingclusterautoscaler.htm#:~:text=If%20you%20are%20using%20both%20Terraform%20and%20the%20Kubernetes%20Cluster%20Autoscaler%20to%20manage%20node%20pools) suggests adding the ignore_changes lifecycle rule for the node_config_details.size field, so it would be very useful to have this provided as an option here.

oci_containerengine_node_pool

robo-cap:
Ability to associate nodepools with `oci_containerengine_node_pool `terraform resources that ignore nodepool size. This is useful for nodepools managed by the cluster autoscaler.

Add `ignore_size = true` to the nodepool attributes.

Adding this attribute to an existing nodepool will trigger the recreation of it.

If you intend to set the attribute ignore_size = true for some of the nodepools, you need to use the [terraform state mv](https://developer.hashicorp.com/terraform/cli/commands/state/mv) command.

Documentation was updated, should be available here: https://oracle-terraform-modules.github.io/terraform-oci-oke/guide/extensions_cluster_autoscaler.html

Resolves #763


modules/workers/locals.tf
```diff
@@ -36,6 +36,7 @@ locals {
     eviction_grace_duration      = 300
     force_node_delete            = true
     extended_metadata            = {} # empty pool-specific default
+    ignore_size                  = false
```
hyder, line 39: You are hardcoding this value. Shouldn't you have a variable to allow the user to choose?

hyder: something like pool-scaling-mode = autoscaler, terraform, manual

Even though manual is similar to non-autoscale, from a user perspective, it's better I think to indicate one as such. This then lets scale terraform-managed workers through terraform only

modules/workers/nodepools.tf
```diff
@@ -156,3 +156,158 @@ resource "oci_containerengine_node_pool" "workers" {
     }
   }
 }
+
+resource "oci_containerengine_node_pool" "with_size_ignored" {
```
hyder, line 160: Please rename this to something that has nodepool in it.

hyder (changes requested):
comments below

modules/workers/locals.tf
robo-cap: I'm using nodepool defaults to set this parameter to avoid the impact on the existing customers.
At the same time, we need to have this control at the nodepool level, we can't use global variables.

modules/workers/nodepools.tf
robo-cap: renamed it to `workers_with_size_ignored` to be consistent with the name of the existing nodepool - `workers` (where the size is not ignored).

modules/workers/instancepools.tf
```diff
@@ -61,3 +61,63 @@ resource "oci_core_instance_pool" "workers" {
     }
   }
 }
+
+resource "oci_core_instance_pool" "autoscaled_workers" {
+  # Create an OCI Instance Pool resource for each enabled entry of the worker_pools map with that mode.
+  for_each                  = { for key, value in local.enabled_instance_pools: key => value if lookup(value, "ignore_size", false) == true }
```
devoncrouse, line 67: Might play it safe and wrap the lookup with `tobool` here and line 7 - values found in maps can be cast to string on you and unexpectedly fail equality checks.

modules/workers/instancepools.tf
```diff
@@ -61,3 +61,63 @@ resource "oci_core_instance_pool" "workers" {
...
+      EOT
+    }
+
+    precondition {
+      condition     = var.cni_type == "flannel"
+      error_message = "Instance Pools require a cluster with `cni_type = flannel`."
+    }
+
+    precondition {
+      condition     = each.value.autoscale == false
```
devoncrouse, line 119: We should remove this precondition - user could be configuring CA on their own/through add-on and still need the ability to `ignore_size`

modules/workers/nodepools.tf
```diff
@@ -2,9 +2,9 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl
 
 # Dynamic resource block for Node Pool groups defined in worker_pools
-resource "oci_containerengine_node_pool" "workers" {
+resource "oci_containerengine_node_pool" "tfscaled_workers" {
   # Create an OKE node pool resource for each enabled entry of the worker_pools map with that mode.
-  for_each           = local.enabled_node_pools
+  for_each           = { for key, value in local.enabled_node_pools: key => value if lookup(value, "ignore_size", false) == false }
```
devoncrouse, line 7: Wrap lookup w/ `tobool` as with instancepools comment, and line 162

docs/src/guide/extensions_cluster_autoscaler.md
```diff
@@ -13,6 +13,66 @@ The following parameters may be added on each pool definition to enable manageme
...
+
+```
+worker_pools = {
+  np-autoscaled = {
+    description = "Node pool managed by cluster autoscaler",
+    size        = 2,
+    min_size    = 1,
+    max_size    = 3,
+    autoscale   = true,
+    ignore_size = true # allow nodepool size drift
```
hyder, line 30: Is allowing drift the default? If yes, please indicate

docs/src/guide/extensions_cluster_autoscaler.md
```diff
@@ -13,6 +13,66 @@ The following parameters may be added on each pool definition to enable manageme
...
+  },
+  np-autoscaler = {
+    description      = "Node pool with cluster autoscaler scheduling allowed",
+    size             = 1,
+    allow_autoscaler = true,
```
hyder, line 35: As above, if I don't set the `ignore_size`, will it default to `true` and allow drift or `false` and not allow drift?

docs/src/guide/extensions_cluster_autoscaler.md
hyder: I think we need to be a bit more explicit and document the various scenarios in a page dedicated to cluster autoscaler.

modules/workers/instancepools.tf
hyder: @devoncrouse If users want to install/configure CA on their own, then they should just set the `ignore_size` variable to `true` and `autoscale` here to `false`. The value of `autoscale` here only impacts the generated autoscaler template to let it know which nodepool/instancepool it should manage. Not whether the nodepools/instancepools themselves.

Am I missing something?

docs/src/guide/extensions_cluster_autoscaler.md
hyder: Come to think of it, `ignore_size` lends to confusion. Let's rename the variable to `ignore_initial_pool_size`.

hyder (dismissed):
comments inline

modules/workers/instancepools.tf
devoncrouse: @hyder You're right, that should be fine - thanks

docs/src/guide/extensions_cluster_autoscaler.md
robo-cap: When it's not set, defaults to `false` and is not allowing configuration drift. (in line with the old behavior).

hyder: ok, please document this in a separate page for autoscaler.

robo-cap: I've update the docs and mentioned the new feature in two places:
- nodepool parameters - because it's a parameter relevant for the nodepools
- cluster autoscaler - because it makes sense to be set when cluster-autoscaler extension is used.

devoncrouse (approved):

Tarasovych:
Thank you for this change!

PR #941, opened 2024-07-04
fix: install mdbook-admonish

hyder:


.github/workflows/mdbook.yaml
```diff
@@ -68,7 +68,14 @@ jobs:
         with:
           key: tfdocs
           path: ./tfdocs/terraform-docs
-
+      - name: Install mdbook-admonish
+        shell: bash
+        env:
+          CARGO_HOME: ./cargo
+        run: |
+          export PATH=`pwd`/cargo/bin:${PATH}
```
Djelibeybi, line 76: I would recommend using `$PWD` instead of `pwd` so you don't invoke a subshell.

Djelibeybi (approved):

PR #948, opened 2024-10-18
add cluster addon support

robo-cap:


docs/src/guide/cluster_addons.md
```diff
@@ -0,0 +1,16 @@
+# Cluster Add-ons
+
+When using enhanced clusters, you can use Kubernetes Engine to manage both essential add-ons and a growing portfolio of optional add-ons.
```
hyder, line 3: Use OKE instead of Kubernetes Engine

hyder (commented):
lgtm. please add a note on cluster autoscaler addon mutual exclusivity with the self-managed cluster autoscaler. we can merge after that

hyder (approved):
approved

PR #955, opened 2024-10-29
Cilium extension upgrade to 1.16

robo-cap:
add support for cilium 1.16.

### DIFF
--- a/docs/src/guide/extensions_networking.md
+++ b/docs/src/guide/extensions_networking.md
@@ -23,6 +23,91 @@
 {{#include ../../../examples/extensions/vars-extensions-cilium.auto.tfvars:4:}}
 ```
 
+Cillium is a eBPF based CNI for Kubernetes that can be configured on OKE clusters.
+
+The OKE cluster should be initially configured to run **flannel**. 
+
+On **enhanced** clusters we can use the cluster-addons module to remove flannel extension and kube-proxy (Optional) at cluster creation. 
+
+
+```
+cluster_addons_to_remove = {
+  Flannel = {
+    remove_k8s_resources = true
+  },
+  KubeProxy = {
+    remove_k8s_resources = true
+  }
+}
+```
+
+If you want to use cilium as [kube-proxy replacement](https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/), you can use the following helm_values:
+
+```
+cilium_helm_values      = {
+  kubeProxyReplacement = true
+}
+```
+
+For the basic clusters you can add the following label to the worker nodes to prevent flannel pods from being scheduled:
+
+```
+oci.oraclecloud.com/custom-k8s-networking=true
+```
+
+If you want to override and of the default values(listed below) you can use the `cilium_helm_values` variable:
+
+```
+"annotateK8sNode": true
+"cluster":
+  "id": 1
+  "name": "oke-${var.state_id}"
+"clustermesh":
+  "apiserver":
+    "kvstoremesh":
+      "enabled": false
+  "useAPIServer": false
+"cni":
+  "exclusive": true
+  "install": true
+"hubble":
+  "metrics":
+    "dashboards":
+      "enabled": false
+  "relay":
+    "enabled": true
+  "ui":
+    "enabled": true
+"installNoConntrackIptablesRules": false
+"ipam":
+  "mode": "kubernetes"
+"k8s":
+  "requireIPv4PodCIDR": true
+"k8sServiceHost": "${var.cluster_private_endpoint}"
+"k8sServicePort": "6443"
+"kubeProxyReplacement": false
+"operator":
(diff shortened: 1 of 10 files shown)


modules/extensions/cilium.tf
```diff
@@ -2,45 +2,47 @@
...
-    exclusive    = !var.multus_install
-  }
-
   cilium_helm_values = {
     annotateK8sNode                 = true
-    cluster                         = { name = "oke-${var.state_id}" }
-    clustermesh                     = { useAPIServer = true }
+    cluster                         = {
+      name = "oke-${var.state_id}"
+      id   = 1
```
hyder, line 26: cluster id should be parameterized instead of hardcoded.

modules/extensions/cilium.tf
```diff
@@ -2,45 +2,47 @@
...
+    }
+    clustermesh                     = { 
+      useAPIServer = false
```
hyder, line 29: Same thing for clustermesh

modules/extensions/cilium.tf
```diff
@@ -2,45 +2,47 @@
...
+        kvstoremesh = {
+          enabled = false
+        }
+      }
+    }
     cni                             = local.cilium_vxlan_cni
-    containerRuntime                = { integration = "crio" }
-    installIptablesRules            = true
     installNoConntrackIptablesRules = false
     ipam                            = { mode = "kubernetes" }
```
hyder, line 38: same thing for ipam mode. set the default to "kubernetes" though

modules/extensions/cilium.tf
```diff
@@ -52,19 +54,9 @@ locals {
 
     k8s = {
       requireIPv4PodCIDR   = true # wait for Kubernetes to provide the PodCIDR (ipam kubernetes)
```
hyder, line 56: In the Cilium docs (https://docs.cilium.io/en/stable/network/concepts/ipam/kubernetes/), they mention the following:

"ipam: kubernetes: Enables Kubernetes IPAM mode. Enabling this option will automatically enable k8s-require-ipv4-pod-cidr if enable-ipv4 is true and k8s-require-ipv6-pod-cidr if enable-ipv6 is true."

Should we consider setting `enable-ipv4` to true?

docs/src/guide/extensions_networking.md
```diff
@@ -23,6 +23,28 @@
 {{#include ../../../examples/extensions/vars-extensions-cilium.auto.tfvars:4:}}
 ```
 
+If you are looking to replace flannel with cilium, you can use the cluster-addons module to remove flannel extension and kube-proxy (Optional).
```
hyder, line 26: Since Cilium helm values have changed recently, add a note to say from which versions of Kubernetes and Cilium this has been tested.

modules/extensions/cilium.tf
robo-cap: This is just a default value I used for deployment simplicity. It can be overridden using the `cilium_helm_values` variable.

modules/extensions/cilium.tf
robo-cap: This can be overridden the same way.

modules/extensions/cilium.tf
robo-cap: can be overridden using the `cilium_helm_values` variable.

hyder (approved):
approved

tcrowder-koerber:
This appears to be a breaking change from 5.1.8 to 5.2.1

module.oke.module.extensions[0].null_resource.cilium[0] (remote-exec): Error: rendered manifests contain a resource that already exists. Unable to continue with install: ServiceAccount "cilium" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "cilium"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"

│   with module.oke.module.extensions[0].null_resource.cilium[0],
│   on .terraform/modules/oke/modules/extensions/cilium.tf line 211, in resource "null_resource" "cilium":
│  211:   provisioner "remote-exec" {
│

robo-cap:
In version 5.2.1 we are deploying cilium using helm. Previously we were using the helm provider only to render the manifest and applied it using kubectl.

tcrowder-koerber:
Is there an upgrade path?

robo-cap:
There is a need to import the existing cilium deployment into a helm chart deployment.

This should be possible with this: [https://github.com/jzbruno/helm-import](https://github.com/jzbruno/helm-import)

Once this is done, to manage everything through terraform, there is a need to extract the values of the current deployment and to apply them to via the `cilium_helm_values` variable, as the current default ones are applicable to cilium 1.16.

PR #962, opened 2024-11-18
feat: add support to reference module nsgs in the nsg rules

robo-cap:
Below is an example of how this can be used.

```
allow_rules_workers = {
    "Allow UDP ingress to workers for port range 50000-52767 from Pub-LB" : {
      protocol = 17, destination_port_min = 50000, destination_port_max=52767, source = "ocid1.networksecuritygroup.oc1.eu-frankfurt-1.aaaaaaaai6z4le2ji7dkpmuwff4525b734wrjlifjqkrzlr5qctgxdsyoyra", source_type = "NETWORK_SECURITY_GROUP"
    },
  }

allow_rules_public_lb = {
    "Allow UDP egress to workers port range 50000-52767 from Pub-LB" : {
      protocol = 17, destination_port_min = 50000, destination_port_max=52767, destination = "workers", destination_type = "NETWORK_SECURITY_GROUP"
    },
  }
```

The supported nsgs to be referenced are: `bastion`, `cp`, `int_lb`, `pub_lb`, `workers`, `operator`, `fss` (when these are created).


hyder:
If I understand this correctly, you are adding the custom rule and applying it in a 2nd terraform run. Is that correct?

hyder (approved):
approved

PR #967, opened 2024-12-08
feat: added stern, update istio-mc examples module

### LINKED ISSUE #964: Add stern to utilities on operator
### Community Note

### Description

[Stern](https://github.com/stern/stern) allows you to tail multiple pods on Kubernetes and multiple containers within the pod. Each result is color coded for quicker debugging.

The query is a regular expression or a Kubernetes resource in the form <resource>/<name> so the pod name can easily be filtered and you don't need to specify the exact id (for instance omitting the deployment id). If a pod is deleted it gets removed from tail and if a new pod is added it automatically gets tailed.

When a pod contains multiple containers Stern can tail all of them too without having to do this manually for each one. Simply specify the container flag to limit what containers to show. By default all containers are listened to.

hyder:
Resolves #964

### DIFF
--- a/variables-operator.tf
+++ b/variables-operator.tf
@@ -89,6 +89,12 @@ variable "operator_install_kubectx" {
   type        = bool
 }
 
+variable "operator_install_stern" {
+  default = true
+  description = "Whether to install stern on the created operator host. NOTE: Provided only as a convenience and not supported by or sourced from Oracle - use at your own risk."
+  type = bool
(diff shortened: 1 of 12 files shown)


variables-operator.tf
robo-cap, line 1: Setting this by default as true won't trigger the replacement of the operator host?

robo-cap (approved):

Issue #969, opened 2024-12-10
Block volume provisioning customization in instance-pool mode


ddevadat:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Block volume provisioning customization in instance-pool mode

### New or Affected Resource(s)

When we create worker node in instance-pool mode, the module creates a 50Gb Block volume and attaches the compute instance.

Right now there is no option to opt out of this. I have seen use case where customer want to create worker pool in instance-pool mode, but dont really need the block volume. Its an extra unwanted cost to them and they cant opt out.

Also if a customer do want to opt in, it would be better to customize the block volume size. Just a thought.

### Potential Terraform Configuration

https://github.com/oracle-terraform-modules/terraform-oci-oke/blob/2c7037a41cc4bea47e648e2f622d22bae2db9f40/modules/workers/instanceconfig.tf#L119C1-L132C6

```hcl
# Copy-paste any Terraform configurations for how the requested feature may be used. 
    block_volumes {
      attach_details {
        type                                = each.value.block_volume_type
        is_pv_encryption_in_transit_enabled = each.value.pv_transit_encryption
      }

      create_details {
        // Limit to first candidate placement AD for cluster-network; undefined for all otherwise
        availability_domain = each.value.mode == "cluster-network" ? element(each.value.availability_domains, 1) : null
        compartment_id      = each.value.compartment_id
        display_name        = each.key
        kms_key_id          = each.value.volume_kms_key_id
      }
    }
```

PR #971, opened 2024-12-11
Feat: Ability to customize block volume for instance-pool

ddevadat:
Added the ability to customize block volume creation when worker pool mode is instance-pool
Fixes issue: [969](https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/969)

Reason for this fix, if customer wants to create thousands of node added as instance-pool, it will end up creating thousands of block volume which customer may or may not need. This creates a potentially unnecessary resource cost.

When worker nodes are added as instance pool, we need an option whether to provision a block volume or not. If we choose to provision a block volume, we should be able to customize the size of the block volume.

Define the worker_pools variable for oke like

```
module "k8s_infra" {
  source                             = "oracle-terraform-modules/oke/oci"
  version                            = "xxxx"
....
worker_pools                       = local.worker_pools
....
}
locals {
  oke-vm-instance-pool-with-block-volume = {
    description = "Self-managed Instance Pool with block volume",
    mode        = "instance-pool",
    size        = 1,
    disable_block_volume = false,
    block_volume_size_in_gbs = 60,
  },
  oke-vm-instance-pool-without-block-volume = {
    description = "Self-managed Instance Pool without block volume",
    mode        = "instance-pool",
    size        = 1,
    disable_block_volume = true,
  },
  }
```

if block_volume_size_in_gbs is not specified the default will be 50gb.

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] New tests
- [ ] Build/CI related changes
- [ ] Documentation content changes
- [ ] Other (please describe)

### DIFF
--- a/examples/workers/vars-workers-instancepool.auto.tfvars
+++ b/examples/workers/vars-workers-instancepool.auto.tfvars
@@ -19,5 +19,18 @@ worker_pools = {
     mode        = "instance-pool",
     size        = 1,
     burst       = "BASELINE_1_8",   # Valid values BASELINE_1_8,BASELINE_1_2
-  }
+  },
+  oke-vm-instance-pool-with-block-volume = {
+    description = "Self-managed Instance Pool with block volume",
+    mode        = "instance-pool",
+    size        = 1,
+    disable_block_volume = false,
+    block_volume_size_in_gbs = 60,
+  },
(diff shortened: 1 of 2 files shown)


hyder (approved):
approved

Issue #974, opened 2024-12-18
Unable to customize CoreDNS configuration when provisioning via oke module


ddevadat:
My requirement is during oke provisioning, i want to customize CoreDNS configuration like change minimum replica count. 
My tfvars for cluster_addon variable looks 
```
  cluster_addons = {
    "CoreDNS" = {
      remove_addon_resources_on_delete = true
      configurations = [
        {
          key   = "minReplica"
          value = "1"
        },
        {
          key   = "nodesPerReplica"
          value = "1"
        }
      ]
    }
  }
```

i am invoking oke like
```

module "oke_blue" {
  source                             = "oracle-terraform-modules/oke/oci" 
  version                            = "5.2.2"
....
....
  cluster_addons          = var.cluster_addons
```

Error received
```
 Error: 409-Conflict, addon CoreDNS exists
 Suggestion: The resource is in a conflicted state. Please retry again or contact support for help with service: Containerengine Addon
 Documentation: https://registry.terraform.io/providers/oracle/oci/latest/docs/resources/containerengine_addon
 API Reference: https://docs.oracle.com/iaas/api/#/en/containerengine/20180222/Cluster/InstallAddon
 Request Target: POST https://containerengine.ap-mumbai-1.oci.oraclecloud.com/20180222/clusters/ocid1.cluster.oc1.ap-mumbai-1.aaaaaaaa/addons
 Provider version: 6.20.0, released on 2024-12-11.
 Service: Containerengine Addon
 Operation Name: InstallAddon
 OPC request ID: 33753b4b49e2eb439c99237106fB9DF718AA4F847A


   with module.oke_blue[0].module.cluster-addons[0].oci_containerengine_addon.secondary_addon["CoreDNS"],
   on .terraform/modules/oke_blue/modules/cluster-addons/addons.tf line 53, in resource "oci_containerengine_addon" "secondary_addon":
   53: resource "oci_containerengine_addon" "secondary_addon" {
```

Potential solution
modules/cluster-addons/addons.tf

Add the flag
override_existing = true

```
resource "oci_containerengine_addon" "primary_addon" {
  for_each = { for k, v in local.addons_with_defaults : k => v if contains(local.primary_addons, k) }

  addon_name = each.key
  cluster_id = var.cluster_id

  remove_addon_resources_on_delete = lookup(each.value, "remove_addon_resources_on_delete", true)

  dynamic "configurations" {
    for_each = lookup(each.value, "configurations", [])
[...]

PR #975, opened 2024-12-18
fix: Ability to override existing cluster addons

ddevadat:
Added the ability to override existing cluster addons via an optional flag
Fixes issue [974](https://github.com/oracle-terraform-modules/terraform-oci-oke/issues/974)


hyder:
This is very useful. Can you please update the docs as well so people can understand how to customize the addon feature? I think you might need to create a new subsection for it.

ddevadat:
> This is very useful. Can you please update the docs as well so people can understand how to customize the addon feature? https://oracle-terraform-modules.github.io/terraform-oci-oke/guide/cluster_addons.html

updated the documentation for customizing an existing addon

hyder (approved):

PR #982, opened 2025-01-08
Fix #980: Add pod NSG to instance configuration metadata #981

### LINKED ISSUE #980: Pod NSGs are not attached to VNICs created by the VCN-native CNI when using self-managed nodes
When creating an instance pool or cluster network with self-managed node using VCN-native CNI for pod networking, pod NSGs are not attached to the secondary VNICs created by the CNI.

OguzPastirmaci:
Resolves #980 

Fix for pod NSGs are not attached to VNICs created by the VCN-native CNI when using self-managed nodes.


hyder (approved):
approved

PR #985, opened 2025-01-28
Improve cluster add-ons documentation

robo-cap:
Provide example about how we can use `nodeSelector` addon configuration and the `override_existing` attribute of the `oci_containerengine_addon` terraform resource, to prevent the pods associated with the essential addons from being scheduled.


hyder (approved):
approved

Issue #989, opened 2025-02-18
cosmetic enhancement :  for each subnet allow to  pass an optional name along with cidr/newbits/id


Umashankar-S:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When  OKE module creates  Subnets the Display Name of Subnets. are cp-<clustername> ,worker-clustername,operator-<clustername> etc ..

### Potential Terraform Configuration

```hcl
module "oke_clus1" {
subnets = {
  bastion  = { newbits = 13, netnum = 0, display_name = "bastion", create="never"  } 
  operator = { cidr = "10.104.17.0/24", display_name = "OKEUATCICD", create="always" }
  cp       = { cidr = "10.104.16.0/24" , display_name = "UATOKECP", create="always" }
  int_lb   = { cidr = "10.104.18.0/23" , display_name = "OKEUATLBPRIV", create="always" }
  pub_lb   = { cidr = "10.104.60.0/25" , display_name = "OKEUATLBPUB", create="never" }
  workers  = { cidr = "10.104.0.0/20",  display_name = "OKEUATWN01", create="always" }
  pods     = { newbits = 2, netnum = 2, display_name = "pods", create="never" }
}


```

PR #993, opened 2025-02-24
Add support to customize subnet name and extend rules for CP and Pods NSGs

robo-cap:
Addresses enhancement #989.


hyder (approved):
approved

PR #1000, opened 2025-04-24
Add IPv6 support

### LINKED ISSUE #996: Support IPV6
For now, the module manages the VCN of the cluster, but it's not providing ability to enable IPV6 on the VCN, although the official VCN module already provides this ability: https://github.com/oracle-terraform-modules/terraform-oci-vcn/blob/f50424b62feb91430870d0e08d67290e82ce87e2/variables.tf#L61C11-L61C22
The OKE module then simply needs to expose another network input which enables/disables IPV6 for the VCN and pass it to the VCN module.

robo-cap:
Resolves #996.


hyder (approved):
approved

PR #1007, opened 2025-05-26
Add compute cluster support

robo-cap:
Add support to deploy RDMA OKE worker nodes in [compute-clusters](https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/compute-clusters.htm).


variables-workers.tf
```diff
@@ -55,6 +55,16 @@ variable "worker_pool_size" {
   type        = number
 }
 
+#
+# Workers: Compute clusters
+#
+
+variable "worker_compute_clusters" {
```
hyder, line 62: Why is this a top-level input variable? The mode should be part of the `worker_pools` I think and it should be able to handle it from there. Am I missing something?

hyder (changes requested):
Please also provide examples

variables-workers.tf
robo-cap: Theoretically, with compute clusters, it's possible to add multiple node shapes to the same RDMA network. Practically, this setup is not recommended nor possible in all the regions (the different shapes must be part of the same block).

The default behavior of the module is to create a separate compute-cluster for each worker-pool configured with `mode = "compute-cluster"`.

For the customers looking to configure worker-pools as part of the same compute-cluster, they can reference an existing compute_cluster using the `compute_cluster` attribute. They can define it via OCID (if created externally) or by name if is created using the `worker_compute_clusters` variable.

robo-cap:
> Please also provide examples

Done.

PR #1032, opened 2025-08-22
Fix default value for `bastion_image_os`

lopter:
`Oracle Autonomous Linux` does not seem to be a thing anymore, and we can use `Oracle Linux` instead.

This makes the "out-of-the-box" experience a little better. But the lack of correct feedback and error propagation for invalid credentials is where multiple days were lost.


lopter:
I am not doing the CLA for this, therefore I moving this to issue #1034, thanks.

PR #1035, opened 2025-09-10
Improve ubuntu worker nodes bootstrapping logic

robo-cap:
Improved ubuntu worker nodes bootstrapping logic.

Added loop to wait for apt lock release before attempt to install the packages.


hyder (commented):
comment in line for discussion

robo-cap:
I created a local map with all the supported ubuntu versions and the corresponding release names.

